PR #19, opened 2026-08-28
fix(translator): correct trait composition precedence and alias modifiers

AlessioGiacobbe:
## Problem

Three defects in `composeTraitAst` diverge from PHP trait semantics:

**1. A concrete method fulfilling another trait's abstract requirement is dropped** (order-dependent):

```php
trait NeedsName { abstract public function name(): string; }
trait HasName { public function name(): string { return "HasName"; } }
class C { use NeedsName, HasName; }   // valid PHP
```

The branch unsets the new *concrete* statement while the already-merged abstract one stays in the class AST, so the compiled class keeps only the abstract declaration → `Cannot instantiate abstract class C` at runtime. `use HasName, NeedsName;` (reverse order) works.

**2. Alias adaptations wipe `static`/`final`/`abstract`:**

```php
trait Maker { public static function make(): string { return "made"; } }
class Factory { use Maker { make as protected; } }
```

`$traitStmt->flags = $alias['newModifier']` replaces the whole flag set with just the visibility bits, so `make` loses `static` and compiles as an instance method. PHP replaces only the visibility bits (verified via Reflection: `make as protected` keeps `isStatic()`, and `m as final` keeps the original visibility).

**3. A class-defined method doesn't suppress trait-vs-trait conflicts:**

```php
trait A { function who() { return "A"; } }
trait B { function who() { return "B"; } }
class C { use A, B; function who() { return "C"; } }   // valid PHP, class wins
```

dies with the spurious fatal ``Trait `B` method `who` already exists`` because the class-method check ran *after* conflict resolution, and suppressed trait copies were still registered as trait methods.

## Fix

- Concrete-fulfills-abstract: drop the already-collected abstract declaration (from the class AST or the pending alias list) and keep the concrete method.
- New `applyTraitAliasModifier()`: a new visibility replaces only `Modifiers::VISIBILITY_MASK` bits; a modifier without visibility keeps the original one.
- The class-own-method check now runs before conflict resolution, and suppressed copies are not registered into `$traitMethods`.

## Tests

- New `tests/compiler/trait/trait-composition-precedence.phpt` covers all three (both trait orders for the abstract case, alias visibility change with same and new name calling through `static::`, class method winning over two conflicting traits). Expected output verified against PHP 8.4.
- Full `tests/compiler/trait/` suite: 41/41 pass. No new PHPStan errors on `Translator.php`.
- On master, the new test's scenarios fail with ``Trait `WhoB` method `who` already exists`` at compile time and, with that scenario removed, `Cannot instantiate abstract class AbstractFirst` at runtime.

### DIFF
--- a/phpunit/code/trait_abstract_alias_incompatible.php
+++ b/phpunit/code/trait_abstract_alias_incompatible.php
@@ -1,0 +1,24 @@
+<?php
+trait RequiresValue
+{
+    abstract public function value(int $value): string;
+}
+
+// Aliasing an abstract method creates the requirement under the new name;
+// the class method defined under that name must satisfy it.
+class AliasImplementation
+{
+    use RequiresValue { value as renamed; }
+
+    public function renamed(string $value): string
+    {
+        return $value;
+    }
+
+    public function value(int $value): string
+    {
+        return "$value";
+    }
+}
+
+function main() {}
--- a/phpunit/code/trait_abstract_class_incompatible.php
+++ b/phpunit/code/trait_abstract_class_incompatible.php
@@ -1,0 +1,17 @@
+<?php
+trait RequiresValue
+{
+    abstract public function value(int $value): string;
+}
+
+class InvalidImplementation
+{
+    use RequiresValue;
+
+    public function value(string $value): string
+    {
+        return $value;
+    }
+}
+
+function main() {}
--- a/phpunit/code/trait_abstract_extra_required_param.php
+++ b/phpunit/code/trait_abstract_extra_required_param.php
@@ -1,0 +1,17 @@
+<?php
+trait RequiresValue
+{
+    abstract public function value(int $value): string;
+}
+
+class GreedyImplementation
+{
+    use RequiresValue;
+
+    public function value(int $value, int $extra): string
+    {
+        return "$value:$extra";
+    }
+}
+
+function main() {}
--- a/phpunit/code/trait_abstract_return_widened.php
+++ b/phpunit/code/trait_abstract_return_widened.php
@@ -1,0 +1,17 @@
+<?php
+trait RequiresValue
+{
+    abstract public function value(): string;
+}
+
+class WideningImplementation
+{
+    use RequiresValue;
+
+    public function value(): string|int
+    {
+        return 'value';
+    }
+}
+
+function main() {}
--- a/phpunit/code/trait_abstract_static_mismatch.php
+++ b/phpunit/code/trait_abstract_static_mismatch.php
@@ -1,0 +1,17 @@
+<?php
+trait RequiresValue
+{
+    abstract public function value(): string;
+}
+
+class StaticImplementation
+{
+    use RequiresValue;
+
+    public static function value(): string
+    {
+        return 'value';
+    }
+}
+
+function main() {}
--- a/phpunit/code/trait_abstract_trait_concrete_first_incompatible.php
+++ b/phpunit/code/trait_abstract_trait_concrete_first_incompatible.php
@@ -1,0 +1,22 @@
+<?php
+trait NeedsName
+{
+    abstract public function name(int $id): string;
+}
+
+trait HasName
+{
+    public function name(string $id): string
+    {
+        return $id;
+    }
+}
+
+// The concrete method is collected first; the later abstract requirement is
+// dropped in its favor but must still be satisfied by it.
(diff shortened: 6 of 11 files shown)


matyhtf (changes requested):
Thank you for the detailed report and implementation. The three original defects are real, and the straightforward cases in the new PHPT are fixed correctly.

I found two remaining trait-semantics issues that need to be addressed before merging. These are implementation issues, not requests for additional test coverage.

1. Abstract trait requirements are discarded without validating the concrete implementation.

The new early class-method check drops every same-name trait method, including abstract requirements:

```php
trait RequiresValue
{
    abstract public function value(int $value): string;
}

class InvalidImplementation
{
    use RequiresValue;

    public function value(string $value): string
    {
        return $value;
    }
}
```

Zend rejects this because the class method is incompatible with the abstract trait method. The PR currently compiles and runs it. The same issue occurs when a concrete method from a later trait replaces an earlier abstract declaration: the abstract declaration is removed without checking the concrete signature.

Before dropping an abstract requirement, please validate the implementation using PHP-compatible method variance rules, including parameter contravariance, return covariance, visibility, static/by-reference/variadic behavior, and required parameters. The existing exact-string comparison used for two abstract declarations should not be reused blindly because a concrete implementation may be valid through variance without being textually identical.

2. Multiple alias adaptations incorrectly depend on source order.

```php
trait AliasSource
{
    public static function value(): string
    {
        return 'value';
    }
}

class AliasConsumer
{
    use AliasSource {
        value as protected;
        value as alias;
    }
}
```

In PHP, `value` becomes protected while `alias` remains public. In the PR, the first adaptation mutates `$traitStmt`, then the second adaptation clones that already-modified node, so `alias` incorrectly becomes protected and cannot be called externally.

Each adaptation should derive its flags independently from the immutable original method flags. A same-name visibility change must not affect the base flags used to create another alias.

I reproduced both differences against Zend PHP 8.4 and by compiling/running the PR head. All current CI jobs pass, but these two cases still diverge from Zend trait composition semantics.

AlessioGiacobbe:
Thanks for the careful review — both issues are addressed in ed65cd2a.

**1. Abstract requirements are now validated before being dropped.** `composeTraitAst` validates the implementation against the abstract declaration in every path that discards a requirement: the class's own method (your `RequiresValue` example), a concrete method from another trait in either collection order (abstract-first and concrete-first), and a class method matching an aliased abstract (`use T { f as g; }` + class-defined `g`, which Zend also validates).

The check follows Zend's trait-composition rules rather than the exact-string comparison used for two abstract declarations: staticness must match, an abstract by-reference return must be kept, the implementation cannot require more parameters, parameter types are contravariant and the return type is covariant (reusing the existing override-check machinery on the preprocessed definitions, with late-bound `self`/`static`/`parent` unified across declaring contexts). One deliberate note on the visibility item in your list: I pinned the behavior against Zend 8.4 first, and Zend places no visibility constraint on the implementation of an abstract trait requirement (`abstract public` + `private` implementation compiles), so visibility is intentionally not restricted — this is documented in the validator.

**2. Alias adaptations no longer depend on source order.** Every adaptation now derives its flags from the immutable original method flags, and the original statement is only mutated after all adaptations are processed. `value as protected; value as alias;` leaves `alias` public and `value` protected in both adaptation orders. Stacked same-name adaptations also match Zend, where each derives from the original and the last one wins (`value as protected; value as final;` yields a final public method).

Coverage added: a new PHPUnit suite (`TraitAbstractRequirementTest`, 7 error cases + a valid-variance case covering contravariant parameters, covariant returns, extra optional parameters, and visibility changes — every code file cross-checked against Zend 8.4), and the composition PHPT now exercises both alias adaptation orders at runtime plus a non-textually-identical fulfillment of an abstract requirement.

matyhtf:
Thank you for the thoughtful revision and the substantial test coverage. The updated implementation resolves the original trait-composition precedence issues, correctly isolates alias modifiers from source order, and adds meaningful validation for abstract trait requirements. This is a strong improvement, and we have merged it.

During the final review we identified two additional late-bound type compatibility edges involving `parent` and explicit trait type names. These are extensions of the compiler's existing trait type-resolution behavior rather than blockers in your core contribution. We will continue refining those two cases on top of this work and add the corresponding positive and negative tests ourselves.

We really appreciate the care you put into validating the behavior against Zend PHP and responding to the review feedback.

PR #28, opened 2026-08-29
fix(optimizer): intval() with a base must not drop the base

### LINKED ISSUE #27: intval() with an explicit base compiles to a cast that drops the base
### Summary

`intval()` is lowered to a single-argument Native cast, so its `$base` is
dropped. `FuncCallOptimizer::dispatchConversion` reads `args[0]` and never looks
at the rest, and the four scalar conversions it serves are modelled as
one-argument casts.

### Reproducer

```php
<?php
function main(): void
{
    echo "intval('ff',16):   ", intval("ff", 16), "\n";
    echo "intval('0x1A',16): ", intval("0x1A", 16), "\n";
    echo "intval('101',2):   ", intval("101", 2), "\n";
}
```

### Expected (PHP 8.5.4)

```
intval('ff',16):   255
intval('0x1A',16): 26
intval('101',2):   5
```

### Actual (compiled binary, PHP 8.5.4 ZTS)

```
intval('ff',16):   0
intval('0x1A',16): 0
intval('101',2):   101
```

The generated C++ shows where it goes:

```cpp
tmp_var_0 = php::toInt(get_str(1));   // intval("ff", 16) - the 16 is gone
```

Nothing warns. The program compiles clean and the number is wrong, which is
unpleasant in the code that most often passes a base: parsing hex colours,
permission masks, binary flags, base36 ids.

### Coverage

`tests/compiler/stdlib/type_conv.phpt` exercises `intval()` with six
single-argument cases; no test anywhere in the suite passes a base. That is why
it went unnoticed.

### Suggested fix

Send a conversion call whose arity is not one to the dynamic path, where both
arguments reach the runtime function. Single-argument `intval()`, `strval()`,
`floatval()` and `boolval()` keep their Native cast. PR follows.

Linux x64. Verified by compiling and running the binary, not only by reading the
generated C++.

Giandonn:
Fixes #27

The four scalar conversions are lowered to a single-argument Native cast by
`dispatchConversion`, which reads `args[0]` and ignores the rest. `intval()`
takes a `$base` as its second argument, so it was silently discarded.

Measured on a compiled binary (PHP 8.5.4 ZTS + embed, GCC 15, Linux x64):

| call | PHP | binary before | binary after |
| --- | --- | --- | --- |
| `intval("ff", 16)` | 255 | 0 | 255 |
| `intval("0x1A", 16)` | 26 | 0 | 26 |
| `intval("101", 2)` | 5 | 101 | 5 |

### The change

A conversion call with any arity other than one now returns `false` from
`dispatchConversion`, so it falls through to the dynamic path and both arguments
reach the runtime function. Single-argument `intval()`, `strval()`, `floatval()`
and `boolval()` still lower to `php::toInt()` and friends, so the common case is
untouched.

### Tests

- `tests/compiler/stdlib/type_conv.phpt` gains the base cases, with a literal
  and a variable base. It already covered six single-argument `intval()` calls;
  no test in the suite passed a base.
- `phpunit/src/ConversionArityTest.php` pins both lowering decisions in the
  generated C++ and needs no native toolchain.

The PHPUnit test fails on master and passes with the change. `type_conv.phpt`
passes against a real build. The full PHPUnit suite reports the same results as
master, and PHPStan reports no new findings for the touched file.

### DIFF
--- a/phpunit/code/intval-single-argument.php
+++ b/phpunit/code/intval-single-argument.php
@@ -1,0 +1,15 @@
+<?php
+/**
+ * This file is part of TypePHP(AOT).
+ *
+ * @link     https://www.swoole.com/aot/
+ * @contact  service@swoole.com
+ */
+
+function main(): void
+{
+    var_dump(intval('42'));
+    var_dump(strval(42));
+    var_dump(floatval('3.14'));
+    var_dump(boolval(1));
+}
--- a/phpunit/code/intval-unpacked-argument.php
+++ b/phpunit/code/intval-unpacked-argument.php
@@ -1,0 +1,21 @@
+<?php
+/**
+ * This file is part of TypePHP(AOT).
+ *
+ * @link     https://www.swoole.com/aot/
+ * @contact  service@swoole.com
+ */
+
+function main(): void
+{
+    $withBase = ['ff', 16];
+    $single = ['42'];
+
+    var_dump(intval(...$withBase));
+    var_dump(intval(...$single));
+    var_dump(strval(...$single));
+    var_dump(floatval(...$single));
+    var_dump(boolval(...$single));
+    var_dump(intval('ff', ...[16]));
+    var_dump(intval(value: '42'));
+}
--- a/phpunit/code/intval-with-base.php
+++ b/phpunit/code/intval-with-base.php
@@ -1,0 +1,15 @@
+<?php
+/**
+ * This file is part of TypePHP(AOT).
+ *
+ * @link     https://www.swoole.com/aot/
+ * @contact  service@swoole.com
+ */
+
+function main(): void
+{
+    $base = 16;
+
+    var_dump(intval('ff', 16));
+    var_dump(intval('ff', $base));
+}
--- a/phpunit/src/ConversionArityTest.php
+++ b/phpunit/src/ConversionArityTest.php
@@ -27,6 +27,21 @@ class ConversionArityTest extends TestCase
         self::assertSame(2, substr_count($cpp, '16L'));
     }
 
+    public function testUnpackedAndNamedArgumentsStayOnTheDynamicPath(): void
+    {
+        $cpp = $this->compileToCpp('intval-unpacked-argument.php');
+
+        // An unpacked argument is one Node\Arg whatever its runtime arity is,
+        // so the array itself must never be handed to a Native cast.
+        self::assertStringNotContainsString('php::toInt(', $cpp);
+        self::assertStringNotContainsString('php::toString(', $cpp);
+        self::assertStringNotContainsString('php::toFloat(', $cpp);
+        self::assertStringNotContainsString('php::toBool(', $cpp);
+
+        // Five full unpacks plus the partial intval('ff', ...[16]).
+        self::assertSame(6, substr_count($cpp, 'appendUnpacked('));
+    }
+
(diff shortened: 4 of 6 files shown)


matyhtf (changes requested):
Thank you for the clear report and for covering both literal and variable bases. The ordinary two-argument `intval()` cases are fixed correctly.

There is still one conversion path that silently drops the call semantics: argument unpacking. An unpacked call has one `Node\Arg` in the AST regardless of its runtime arity, so the new `count($expr->args) === 1` check still accepts it as a single-argument Native conversion.

For example:

```php
$withBase = ['ff', 16];
var_dump(intval(...$withBase)); // PHP: 255

$single = ['42'];
var_dump(intval(...$single));   // PHP: 42
```

With this PR, both calls are currently lowered as conversions of the array itself:

```cpp
php::toInt(withBase);
php::toInt(single);
```

This produces the wrong result and bypasses unpacking entirely. The same structural issue applies to unpacked `strval()`, `floatval()`, and `boolval()` calls.

Please also reject an unpacked argument in `dispatchConversion()`, for example:

```php
if (
    count($expr->args) !== 1
    || !($expr->args[0] instanceof Node\Arg)
    || $expr->args[0]->unpack
) {
    return false;
}
```

That form must stay on the dynamic call path so the runtime can determine the expanded arity. Please add PHPT coverage for both `intval(...['ff', 16])` and a single expanded argument, plus a code-generation assertion that the unpacked array is not passed to `php::toInt()`.

The PR's current targeted PHPUnit and PHPT tests pass locally, but no GitHub checks are reported yet. Once the unpack case is covered, the core fix looks good.

Giandonn:
Thank you — you are right, and the failure is worse than a wrong result. I have pushed the guard.

**Reproducing it.** Adding your two examples to `type_conv.phpt` and running them against the previous version of this PR:

```
017- int(255)          017+ int(1)
018- int(42)           018+ int(1)
019- string(2) "42"    019+ Warning: Array to string conversion ... string(5) "Array"
020- float(42)         020+ float(1)
```

So `intval(...['ff', 16])` was counting the array rather than converting it, and `strval(...)` stringified the array outright.

**The guard.** `dispatchConversion()` now rejects the unpacked form as you suggested. I also added `$expr->args[0]->name !== null`, because a named argument has exactly the same shape and need not be the value being converted:

```php
intval(bogus: 1);   // PHP: Error: Unknown named parameter $bogus
                    // before: php::toInt(1) -> int(1)
```

That makes the four conversions agree with `dispatchFuncCall()`, which already rejects both `$arg->name !== null` and `$arg->unpack` for every other builtin. If you would rather keep this PR to the unpack case alone, say the word and I will split the named-argument part out.

**Coverage.** `phpunit/code/intval-unpacked-argument.php` holds the six unpacked forms plus the named one; the code-generation test asserts that none of `php::toInt(` / `php::toString(` / `php::toFloat(` / `php::toBool(` appears and that all six unpacks lower to `appendUnpacked(`. `type_conv.phpt` covers the same calls at runtime, including `intval(...['ff', 16]) === 255`, the single expanded argument, and the partial `intval('ff', ...[16])`.

Verified locally on PHP 8.5.4 ZTS with embed: the PHPT compiles and runs green, `ConversionArityTest` passes, PHPStan is clean on the touched file, and the full PHPUnit run matches master.

PR #32, opened 2026-08-30
fix(generator): float constants must not be truncated by the precision ini

### LINKED ISSUE #31: Float constants are truncated to the host's precision ini and baked into the binary
### Summary

`Generator\Utils::genCValue()` lowers a float with a string cast:

```php
if (is_float($value)) {
    return (string) $value;
}
```

A string cast formats using the `precision` ini, which **defaults to 14**. Every
float constant the compiler resolves is therefore truncated to 14 significant
digits inside the generated C++, and the wrong value is baked into the binary.

### Reproducer

```php
<?php
function main(): void
{
    var_dump(M_E);
    var_dump(log(M_E));
}
```

Compiled by a host with the default `precision=14`, the generated C++ contains:

```cpp
php::Var e = 2.718281828459;   // M_E, truncated
php::fn::log(2.718281828459);
```

so the binary prints `float(0.9999999999999832)` where PHP prints `float(1)`.

Compile the same file with `php -d precision=17 bin/tpc.php` and the constant
comes out as `2.7182818284590451` and the binary agrees with PHP. The output of
the compiler depends on how the host PHP happens to be configured.

### This already breaks three tests in the suite

`run-tests.php` sets `precision=14` itself (line 303), so on a checkout with no
precision override these fail on master:

- `tests/compiler/basic/math_functions.phpt` - `log(M_E)` gives
  `float(0.9999999999999832)` instead of `float(1)`
- `tests/compiler/class/readonly-class.phpt` - `M_PI * 5 ** 2` gives
  `float(78.539816339745)` instead of `float(78.53981633974483)`
- `tests/compiler/type_decl/union-intersection-types.phpt` - same `M_PI` case

They pass in CI because the workflow sets `precision=17` in its ini values.

### Fix

`BinaryOpTrait::genFloatLiteral()` already does this correctly - `%.17g` plus a
guard that keeps the literal a C++ double - and `genCValue()` simply does not
call it. Right now the same double is spelled two different ways depending on
which path emits it. PR follows; all three tests above go green.

Linux x64.

Giandonn:
Fixes #31

`genCValue()` lowered a float with a string cast, which formats using the
`precision` ini. That defaults to 14, so a constant baked into the binary lost
digits whenever the host PHP was not configured otherwise:

```cpp
// compiled by a host with the default precision=14
php::Var e = 2.718281828459;   // M_E
php::fn::log(2.718281828459);  // 0.9999999999999832, PHP gives 1
```

The value is wrong in the binary itself, so nothing at runtime recovers it, and
the same source compiled on two differently configured hosts produces two
different programs.

### The change

`BinaryOpTrait::genFloatLiteral()` already formats with `%.17g` and keeps the
literal a C++ double. `genCValue()` was simply not using it, so the same double
was spelled two ways depending on which path emitted it. One line, and both
paths now agree.

### Three existing tests go green

`run-tests.php` sets `precision=14` itself, so these fail on master on a
checkout without a precision override, and pass with this change:

```
tests/compiler/basic/math_functions.phpt
tests/compiler/class/readonly-class.phpt
tests/compiler/type_decl/union-intersection-types.phpt
```

Verified against a real build: PHP 8.5.4 ZTS with embed, GCC 15, Linux x64.
They pass in CI today only because the workflow sets `precision=17`.

### Two unit tests updated, on purpose

`UtilsTest::testGenCValueFloat` and
`CompilerBaseApiTest::testGeneratedCValuesAreAlwaysSourceCodeStrings` asserted
the output by comparing against `(string) $value` - the precision-dependent cast
itself, so they pinned the behaviour rather than the requirement. They now
assert that the emitted literal reads back as the same double, which is what the
generated code actually needs. `FloatLiteralPrecisionTest` covers the regression
head-on by compiling with `precision=14` set.

The rest of the PHPUnit suite reports the same results as master, and PHPStan
reports no findings for the touched file.

### DIFF
--- a/phpunit/code/float-constant-precision.php
+++ b/phpunit/code/float-constant-precision.php
@@ -1,0 +1,13 @@
+<?php
+/**
+ * This file is part of TypePHP(AOT).
+ *
+ * @link     https://www.swoole.com/aot/
+ * @contact  service@swoole.com
+ */
+
+function main(): void
+{
+    var_dump(M_E);
+    var_dump(log(M_E));
+}
--- a/phpunit/src/CompilerBaseApiTest.php
+++ b/phpunit/src/CompilerBaseApiTest.php
@@ -361,7 +361,9 @@ PHP);
 
     public function testGeneratedCValuesAreAlwaysSourceCodeStrings(): void
     {
-        $this->assertSame((string) M_E, $this->invokeMethod('genCValue', M_E));
+        // A string cast here would follow the precision ini, so the assertion
+        // is that the emitted literal reads back as the same double.
+        $this->assertSame(M_E, (float) $this->invokeMethod('genCValue', M_E));
         $this->assertSame('1', $this->invokeMethod('genCValue', true));
         $this->assertSame('0', $this->invokeMethod('genCValue', false));
 
@@ -370,7 +372,7 @@ PHP);
             new \PhpParser\Node\Expr\ConstFetch(new \PhpParser\Node\Name('M_E'))
         );
         $this->assertIsString($code);
-        $this->assertSame((string) M_E, $code);
+        $this->assertSame(M_E, (float) $code);
     }
 
     public function testNumericStringIdentifiersGenerateSourceCodeStrings(): void
--- a/phpunit/src/FloatLiteralPrecisionTest.php
+++ b/phpunit/src/FloatLiteralPrecisionTest.php
@@ -1,0 +1,50 @@
+<?php
+/**
+ * This file is part of TypePHP(AOT).
+ *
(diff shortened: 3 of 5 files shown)


matyhtf:
Thank you for the precise diagnosis of the host precision issue and for adding focused regression coverage. Your report correctly identified the root cause in the float code-generation path. PR #33 has now been merged with a consolidated implementation that includes the same precision fix and additionally covers locale-independent formatting, INF/NAN values, and the persistent declaration path used by class constants and property defaults. Because the two changes overlap, we are closing this PR as superseded rather than merging both. We sincerely appreciate your contribution.

PR #33, opened 2026-08-30
fix(generator): normalize float literal emission and handle INF/NAN constants

### LINKED ISSUE #31: Float constants are truncated to the host's precision ini and baked into the binary

### LINKED ISSUE #34: Constant expressions evaluating to INF, -INF, or NAN generate invalid C++ syntax
### Summary

When constant-folding expressions or resolving float values that evaluate to `INF`, `-INF`, or `NAN`, the compiler emits `"INF"`, `"-INF"`, and `"NaN"` / `"NAN"` as raw tokens in generated C++ files instead of `std::numeric_limits<double>` calls.

In C++, `INF` and `NaN` are not valid floating-point literals, causing C++ compilation of the generated translation units to fail.

### Reproducer

```php
<?php

function get_special(): array
{
    return [
        INF,
        -INF,
        NAN,
    ];
}
```

Generated C++ output:
```cpp
// Invalid tokens emitted:
php::Var a = INF;
php::Var b = -INF;
php::Var c = NaN;
```

### Expected Behavior

Special float values should be lowered as:
- `std::numeric_limits<double>::infinity()` for `INF`
- `-std::numeric_limits<double>::infinity()` for `-INF`
- `std::numeric_limits<double>::quiet_NaN()` for `NAN`

consistent with the definitions used elsewhere in the compiler.

prateekbhujel:
### Summary

Consolidates float literal generation into `Utils::genFloatLiteral()` so all float-to-C++ codegen paths handle precision, locale independence, and special values consistently.

Fixes #31
Fixes #34

### Details

1. `genCValue()` and `gen_stub.php` (`EvaluatedValue::getCExpr()`) previously stringified floats directly, which:
   - Truncated float constants to the host's `precision` ini (default 14) instead of full 17 digits (#31).
   - Emitted `INF` / `-INF` / `NAN` as raw invalid C++ tokens in expressions, class constants, and property defaults (#34).
   - Emitted whole numbers like `1.0` as `"1"`, losing floating-point type identity in C++.
2. Switched from `%g` to PHP's locale-independent `%h` specifier (`%.17h`) to avoid comma decimal separators under European locales.
3. `genFloatLiteral()` is now the single public helper under `TypePhp\Generator\Utils`, mapping special values to `std::numeric_limits<double>` and preserving `.0` on whole numbers.
4. Removed dead `$floatPrecision` property from `CompilerBase`.
5. Added test coverage in `OperatorTest` for runtime expressions, class constant/property declaration metadata, and locale independence.

### DIFF
--- a/phpunit/code/float-declaration-metadata.php
+++ b/phpunit/code/float-declaration-metadata.php
@@ -1,0 +1,19 @@
+<?php
+
+class FloatDeclarationContainer
+{
+    public const float POSITIVE_INF = INF;
+    public const float NEGATIVE_INF = -INF;
+    public const float NOT_A_NUMBER = NAN;
+    public const float CONST_E = M_E;
+    public const float CONST_ONE_POINT_FIVE = 1.5;
+
+    public float $property_e = M_E;
+    public float $property_inf = INF;
+    public float $property_nan = NAN;
+    public float $property_one_point_five = 1.5;
+}
+
+function main(): void
+{
+}
--- a/phpunit/code/float-literal-special.php
+++ b/phpunit/code/float-literal-special.php
@@ -1,0 +1,13 @@
+<?php
+
+function get_special_floats(): array
+{
+    return [
+        1.0,
+        0.0,
+        INF,
+        -INF,
+        NAN,
+        M_E,
+    ];
+}
--- a/phpunit/src/CompilerBaseApiTest.php
+++ b/phpunit/src/CompilerBaseApiTest.php
@@ -361,7 +361,7 @@ PHP);
 
     public function testGeneratedCValuesAreAlwaysSourceCodeStrings(): void
     {
-        $this->assertSame((string) M_E, $this->invokeMethod('genCValue', M_E));
+        $this->assertSame(M_E, (float) $this->invokeMethod('genCValue', M_E));
         $this->assertSame('1', $this->invokeMethod('genCValue', true));
         $this->assertSame('0', $this->invokeMethod('genCValue', false));
 
@@ -370,7 +370,7 @@ PHP);
             new \PhpParser\Node\Expr\ConstFetch(new \PhpParser\Node\Name('M_E'))
         );
         $this->assertIsString($code);
-        $this->assertSame((string) M_E, $code);
+        $this->assertSame(M_E, (float) $code);
     }
 
     public function testNumericStringIdentifiersGenerateSourceCodeStrings(): void
--- a/phpunit/src/Generator/UtilsTest.php
+++ b/phpunit/src/Generator/UtilsTest.php
@@ -49,7 +49,8 @@ class UtilsTest extends TestCase
     public function testGenCValueFloat(): void
     {
         $result = $this->invokeMethod('genCValue', 3.14);
-        $this->assertSame((string) 3.14, $result);
+        $this->assertSame(3.14, (float) $result);
+        $this->assertMatchesRegularExpression('/[.E]/i', $result);
     }
 
     public function testGenCValueBool(): void
--- a/phpunit/src/OperatorTest.php
+++ b/phpunit/src/OperatorTest.php
@@ -85,4 +85,81 @@ class OperatorTest extends \BaseTest
     {
         $this->exec('Cannot divide or modulo by zero', 'assign-modulo-by-zero.php');
     }
+
+    public function testFloatLiteralSpecialValuesAndWholeNumbers(): void
+    {
+        $previous = ini_set('precision', '14');
+        try {
+            global $translator;
+            $compiler = \TypePhp\CompilerTest::create(TYPEPHP_ROOT_PATH);
+            $translator = $compiler;
+            $testFile = __DIR__ . '/../code/float-literal-special.php';
+            $compiler->addFiles([$testFile]);
+            $compiler->prepareFile($testFile);
+            $cppFile = $compiler->convertFile($testFile);
+            $cpp = file_get_contents($cppFile);
+        } finally {
+            if ($previous !== false) {
+                ini_set('precision', $previous);
+            }
+        }
+
+        $this->assertStringContainsString('1.0', $cpp);
+        $this->assertStringContainsString('0.0', $cpp);
+        $this->assertStringContainsString('std::numeric_limits<double>::infinity()', $cpp);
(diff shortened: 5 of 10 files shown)


matyhtf (changes requested):
Thank you for consolidating the float literal generation paths. The overall direction is correct, and the new tests cover ordinary runtime expression lowering well.

I found two implementation issues that need to be addressed before merging.

1. The persistent declaration path in `gen_stub.php` still emits floats through `strval()`.

`EvaluatedValue::getCExpr()` currently contains:

```php
} elseif ($this->type->isInt() or $this->type->isFloat()) {
    return strval($this->value);
}
```

As a result, class constants and property defaults still generate raw special-value tokens:

```cpp
ZVAL_DOUBLE(&const_POSITIVE_value, INF);
ZVAL_DOUBLE(&const_NEGATIVE_value, -INF);
ZVAL_DOUBLE(&const_NOT_A_NUMBER_value, NAN);
```

Finite values in the same path also remain dependent on the host `precision` setting. With `precision=14`, `M_E` still generates:

```cpp
ZVAL_DOUBLE(&const_E_value, 2.718281828459);
ZVAL_DOUBLE(&property_e_default_value, 2.718281828459);
```

This means the PR does not yet completely fix either #31 or #34. Please route float values emitted by `EvaluatedValue::getCExpr()` through the same canonical float literal formatter. This may require making the formatter accessible to `gen_stub.php`, or moving it to a shared stateless helper.

2. `sprintf('%.17g', ...)` is locale-dependent and can silently generate incorrect C++.

PHP's `%g` conversion uses the decimal separator from `LC_NUMERIC`. For example, under the available `en_DK.utf8` locale:

```php
sprintf('%.17g', 1.5) // "1,5"
```

The PR then generates:

```cpp
return php::toFloat(1,5.0);
```

This is parsed as a C++ comma expression and evaluates to `5.0`, so it is a semantic miscompile rather than only a syntax issue.

Please use PHP's locale-independent `%h` conversion instead, for example:

```php
$text = sprintf('%.17h', $value);
```

The existing exponent/decimal preservation logic can remain the same.

After consolidation, `CompilerBase::$floatPrecision` also appears to have no remaining readers. Removing it would avoid retaining dead configuration, though this cleanup is not a merge blocker.

I reproduced both blocking cases on the PR head. The current focused PHPUnit tests pass, but they do not exercise declaration metadata or a locale with a comma decimal separator.

matyhtf:
On a personal note, I was deeply saddened to hear about the devastating flash floods and landslides affecting Nepal and communities along the Nepal–China border.

My heartfelt condolences go to everyone who has lost family members, friends, and loved ones. My thoughts are also with those who remain missing, those who have been injured or displaced, and the rescue workers and volunteers facing extraordinarily difficult conditions.

I sincerely hope that you, your family, and your friends are safe. May those who lost their lives rest in peace, and may the affected communities find strength, support, and recovery in the difficult days ahead.

prateekbhujel:
Thank you @matyhtf for the thorough review and catch!

I have addressed all the feedback in the latest update:

1. **Stub declaration path in `gen_stub.php`**: Routed float expressions in `EvaluatedValue::getCExpr()` through `getTranslator()->genFloatLiteral()`. Class constants and persistent property defaults now correctly emit `std::numeric_limits<double>` tokens and full 17-digit precision without host ini dependency.
2. **Locale-independence**: Switched from `%.17g` to PHP's locale-independent `%.17h` format specifier.
3. **Dead code cleanup**: Removed the unused `$floatPrecision` property from `CompilerBase`.
4. **Test coverage**: Added dedicated test cases in `OperatorTest` covering declaration metadata (`_arginfo.h` constants and extension property defaults) and locale independence under comma-decimal locales.

And thank you so much for your kind words and thoughts regarding the situation in Nepal. It means a lot. Everyone in our family is safe. Wishing you all the best as well!

matyhtf (approved):
Thank you for addressing all previous review findings. I verified that the persistent declaration path now uses the canonical formatter, the locale-independent float format is used, and the dead precision configuration has been removed. I also compiled the new declaration fixture with precision=14; the generated C++ builds and runs successfully.

There is a small isolation issue in the new locale test: setlocale returns the newly selected locale rather than the previous one, so the test does not restore the original locale. This does not affect the generator fix, and we will correct the test on master in a follow-up commit. The implementation itself is approved.

prateekbhujel:
Thank you for the merge and the follow-up locale fix, the LC_NUMERIC-only approach with the skip guard is much cleaner than what I had. Appreciate you taking the time to tidy that up.

Really glad this contribution made it in. Looking forward to digging deeper into the codebase and contributing more where I can.

PR #39, opened 2026-08-31
fix: 27 correctness divergences from Zend 8.4 across codegen, folding, and declaration validation

AlessioGiacobbe:
This PR fixes 27 verified correctness divergences from Zend PHP 8.4 semantics, found by systematically differential-testing the compiler front-end against Zend 8.4.13 and inspecting the generated C++/arginfo. Every fix was probed against Zend first (the exact rule, both directions), and every fix ships with PHPUnit coverage; runtime-visible fixes also add phpt tests with Zend-validated expectations.

The findings fall into three groups.

## Miscompilations — valid PHP that compiled but behaved differently at runtime

- **Typed `int`/`float` division emitted raw C++ `/`** — `divInts(7, 2)` returned `3.0` instead of `3.5`; `/ 0` was UB instead of `DivisionByZeroError`; `float / 0.0` produced `INF` instead of throwing. Division now routes through `php::Var` like `+ - *` already did. Both-int `%` (the existing guard was inverted) and dynamic `<<`/`>>` (UB for counts ≥ 64 or negative) are routed the same way; constant folds are untouched.
- **Unary minus did not parenthesize its operand** — `-($a ? $b : $c)` compiled to `-(toBool(a)) ? b : c`, losing the sign and selecting the wrong branch (sibling of #20's `--x` case, which the narrow guard could not catch). The operand is now always parenthesized.
- **`??=` evaluated a side-effecting RHS unconditionally** when the RHS materialized statements. The native-object conditional-lambda mechanism is generalized so the RHS executes only in the not-set branch.
- **Compound assignment and `++`/`--` on typed scalar slots were raw C++** — `$a /= $b` truncated, `+=` overflow wrapped (UB) instead of following the compiler's own established assignment semantics. These now lower to the equivalent plain assignment through the safe binary path.
- **Later-operand side effects were hoisted above earlier reads** — `two($j, $j = 5)` with `$j = 1` printed `"5,5"` (Zend: `"1,5"`); same for string concatenation. Earlier plain-variable operands are now snapshotted at their exact Zend read positions (concat's read positions were probed empirically; `$k + ($k = 5)` codegen is byte-identical, matching Zend's CV-read-at-op behavior).
- **Enum cases in class-constant position folded to their backing value or case-name string** — `K::CB === E::B` compiled to `false`, and expression-valued cases (`case A = 1 + 1;`) warned during compilation and lost their value. Case expressions are now evaluated with the constant evaluator, and enum-case-valued constants are re-bound each RINIT via `php::updateConstant(..., php::getEnumCase(...))` (enum case objects are request-scoped, so a MINIT zval is impossible), mirroring the array-constant mechanism.
- **`PHP_INT_MIN` class constants emitted an ill-formed C literal** (`-9223372036854775808`); now `ZEND_LONG_MIN`. (The precision-14 float emission and `count()` folding issues found in the same audit were already fixed on master by #24/#33 — thanks!)
- **Auto-Decimal literal classification misfired** — exponent digits and trailing zeros counted as significant (`is_float(2.220446049250313E-16)` compiled to `false`), overflowing hex literals containing the digit `E` became `Decimal("0x…")`, and valid `var_export()` round-trip literals like `0.1 + 0.2 == 0.30000000000000004` failed to compile. Classification now counts real mantissa precision with a round-trip-exactness check, excludes hex/octal/binary, and demotes a Decimal-classified bare literal to its exact double when it meets a float-typed expression.
- **The `PHP_INT_MAX`/`PHP_INT_MIN` fold was case-insensitive and namespace-blind** — `namespace N; const PHP_INT_MAX = 5; PHP_INT_MAX + 1` folded to `9.2e18` instead of reading the namespaced constant. Resolution now mirrors `parseConstFetch`.
- **Literal `/ 0` and `% 0` were a compile-time fatal even in dead/guarded code**, unlike the equivalent `1 % (1 - 1)` spelling. The literal spelling now warns and produces the same catchable runtime `DivisionByZeroError`.

## Accepted invalid PHP — Zend compile fatals that typephp registered anyway

Compiled classes are installed as internal classes, where ZendVM does not re-run these checks, so these hierarchies previously reached runtime:

- **Overrides of methods inherited from built-in classes had no signature check** (`class C extends ArrayObject { public function offsetGet(int $key): string }` compiled). A reflection-based validator now enforces the full Zend ruleset — visibility, staticness, by-ref return covariance, parameter contravariance with variadic absorption, return covariance — while exempting *tentative* return types, which Zend only deprecates.
- **`__construct` overrides skipped every check**, including `final` parent constructors and abstract-constructor signature contracts. Ordinary concrete parent constructors remain exempt from LSP variance, as in Zend.
- **Interface constants were never validated** — type covariance, `final` override bans, visibility, and two-origin ambiguity are now enforced through an effective-constants table that preserves the original declaring interface, including contracts arriving via parent classes, parent interfaces, and enum `implements`.
- **Static ↔ instance property override mismatches were accepted** in both directions.
- **Abstract-method redeclarations were never validated** — turning a concrete inherited method abstract compiled, and abstract-over-abstract redeclarations skipped the signature check. Trait-originated abstract requirements remain exempt (Zend lets an inherited concrete method satisfy them).
- **Enum declaration rules were unenforced** — properties, forbidden magic methods, case values on non-backed enums, missing values on backed enums, duplicate case names, non-int/string backing types, explicit `implements UnitEnum`, and `class B extends Enum` all compiled.
- **Readonly declaration rules were unchecked for ZendVM-backed classes** (defaults, missing types, static readonly), and **readonly-class inheritance** was not sealed in either direction.
- **Interface declarations accepted bodies and invalid modifiers**, and `interface I extends SomeClass` was misreported instead of rejected.
- **Same-name methods were never cross-checked when interfaces merge** — `interface J extends I1, I2` (and a class implementing both without defining the method) compiled with mutually incompatible declarations. The first-seen declaration is now validated as an override of every later one, with diamond inheritance and class-defined methods exempt, matching Zend's merge order.
- **Property-hook placement rules** (hooks on static/readonly properties, abstract hooks in concrete classes) were unenforced on the class path.
- **Variadic promoted constructor properties and `callable` property types** were accepted.
- **Trait adaptations referencing nonexistent methods were silently ignored** — `use A { missing as g; }` and `B::f insteadof A` with no `B::f` now fail with Zend's diagnostics (unqualified aliases are grouped so any matching trait satisfies them).
- **Compound-type well-formedness** — duplicate union members, `?mixed`, `mixed|T`, and related invalid declarations.

## Rejected valid PHP

- **Trait constant/property conflicts were compared by pretty-printed source text** — two traits declaring `const int X = 1 + 1;` and `const int X = 2;` (or `[1, 2]` vs `array(1, 2)` defaults) were rejected as conflicting. Values are now compared by evaluation, with Zend's declaration-time int→float coercion for float-typed members.
- **Adding a by-ref return in an override was rejected** — Zend only forbids dropping one.
- **A trailing child variadic absorbing parent parameters was rejected** (the decorator pattern) — now validated per Zend's absorption rules.
- **Typed class constants were forced invariant** — PHP 8.3 constants are covariant; narrowing (`int|string` → `int`) is now accepted, widening still rejected.
- **Redeclaring a parent's private method was rejected** — private methods are not inherited in PHP. Verified safe for devirtualization: private-resolved calls devirtualize to the *declaring* class's body, matching PHP's private-scope binding, and Native classes give private methods no virtual slot.

## Verification

- Every fix was pinned against Zend 8.4.13 behavior before implementation (including message wording and edge-case matrices — e.g. 32 constant-inheritance probes, 13 built-in-override probes, 7 concat evaluation-order shapes).
- A 43-case differential acceptance gate (valid-PHP-must-accept / invalid-PHP-must-reject / generated-C++ pattern checks), all Zend-ground-truthed, passes end to end.
- The full `tests/compiler` corpus (1132 phpt files) front-end-compiles with zero new failures.
- ~130 new PHPUnit tests across 20 suites; every error fixture cross-checked to fatal under real Zend 8.4.13 and every valid fixture to run clean. New phpt tests carry Zend-validated expectations.
- The PHPUnit suite (1729 tests) shows zero new failures against a pristine checkout of the same base — the pre-existing environmental failure set on macOS is identical before/after (one apparent delta, `CompilerBaseApiTest::testLibraryImportStubCombines…`, turned out to reproduce on pristine master too and is repo-path-dependent, not caused by this PR).

Found while working on #19 (this PR is independent of it, though both touch trait composition — whichever merges second may need a trivial rebase). The audit also confirmed #20/#21/#24/#33 fixed their targets. Happy to split this into per-subsystem PRs if that's easier to review.


matyhtf:
Thank you very much for the substantial effort and contribution in this PR. We really appreciate the systematic investigation, the detailed explanations, and the large amount of accompanying test coverage.

However, this PR is exceptionally large and contains many independent behavioral changes and fixes across code generation, constant folding, declaration validation, inheritance, traits, enums, and the test suite. Reviewing these changes safely requires a significant amount of time and careful verification against both Zend PHP semantics and TypePHP's existing design constraints. For that reason, we cannot merge the PR directly as a single change set.

We will continue reviewing it incrementally. We may split or extract smaller, well-scoped parts of the implementation and merge them into `master` step by step after each part has been independently reviewed and verified. We will preserve the appropriate authorship and credit for the contributed work.

Thank you again for your patience and for investing so much effort in improving TypePHP.

matyhtf:
Thank you again for the exceptionally thorough audit, detailed analysis, and extensive test coverage.

After reviewing the scope, we have decided to treat PR #39 as a reference source for verified issues and tests, rather than as a branch to be merged directly. The changes span too many independent compiler subsystems and need to be reviewed and integrated separately to preserve correctness and make regressions easier to isolate.

We plan to split the findings into approximately 18 independent issues and address them one by one through small, focused PRs. We will continue using the reproductions, Zend comparisons, tests, and implementation ideas from this PR as important reference material, with appropriate credit retained.

Therefore, PR #39 itself will not be merged. We will close it later after the individual work items have been recorded and organized. Thank you for your understanding and for the substantial amount of work you contributed.

AlessioGiacobbe:
> Thank you again for the exceptionally thorough audit, detailed analysis, and extensive test coverage.
> 
> After reviewing the scope, we have decided to treat PR #39 as a reference source for verified issues and tests, rather than as a branch to be merged directly. The changes span too many independent compiler subsystems and need to be reviewed and integrated separately to preserve correctness and make regressions easier to isolate.
> 
> We plan to split the findings into approximately 18 independent issues and address them one by one through small, focused PRs. We will continue using the reproductions, Zend comparisons, tests, and implementation ideas from this PR as important reference material, with appropriate credit retained.
> 
> Therefore, PR #39 itself will not be merged. We will close it later after the individual work items have been recorded and organized. Thank you for your understanding and for the substantial amount of work you contributed.

i would be happy to split this into smaller, scoped, issues if that's ok with you and if i don't overlap with somebody else's work, let me know if i can contribute that way

matyhtf:
Your contribution is very welcome — thank you.

Please split this into smaller pull requests, one issue per PR. This one is +5,423 / −146 across 27 distinct fixes, which is too large for us to review carefully as a single change. Smaller PRs are much easier to review, and we can merge the straightforward ones immediately instead of holding all 27 behind a single review cycle.

The three groups in your description are a reasonable starting point, and each can be broken down further by individual finding.

Please also rebase onto the latest `master` to avoid conflicts.

AlessioGiacobbe:
@matyhtf 

Done — split into 22 single-issue PRs, each rebased onto the latest master, self-contained (fix + its tests), and re-verified individually. Closing this one in favor of:

**Codegen / folding**
- #45 division, modulo, shifts and compound assignment on typed scalars (incl. literal `/ 0` becoming the catchable runtime error)
- #46 unary-minus operand parenthesization (ternary/`?:` case)
- #47 `??=` lazy evaluation of compound RHS
- #48 `PHP_INT_MIN` constants as `ZEND_LONG_MIN`
- #49 namespace-aware, case-sensitive `PHP_INT_MAX`/`MIN` folds
- #50 auto-Decimal literal classification
- #51 enum-case class constants registered as real case objects
- #52 Zend operand read order around hoisted side effects

**Translator validation**
- #53 overrides may add a by-ref return
- #54 trailing child variadic absorbing parent parameters
- #55 signature validation for overrides of built-in class methods
- #56 constructor / private-method / abstract-redeclaration override rules
- #57 static ↔ instance property redeclaration
- #58 covariant typed class constants + interface constant contracts
- #59 interface declaration and merge rules
- #60 trait member value comparison + adaptation validation

**Preprocessor declaration rules**
- #61 enum declaration rules
- #62 abstract method bodies / abstract private methods
- #63 property-hook placement
- #64 variadic promoted properties, callable property/constant types
- #65 compound type declarations
- #66 readonly declaration and inheritance rules

They are independent of each other (each is green against master on its own); where an upstream change already covered part of a finding — count() folding, float-literal precision, typed-property compound assignment (af466b06) — the split PRs contain only the residual and say so. Happy to split any of these further if one still feels too large.

PR #51, opened 2026-09-01
fix(codegen): register enum-case class constants as real case objects

AlessioGiacobbe:
A class constant valued by an enum case was registered as the case's backing value (or its name string for pure enums): `enum E: int { case B = 4; } class K { const CB = E::B; }` registered `ZVAL_LONG(4)`, so `K::CB === E::B` was `false` — observable through `constant()`, `$cls::CB`, and reflection. Expression-valued cases (`case A = 1 + 1;`) also warned "Undefined property …::$value" during compilation and lost their value in compile-time metadata.

Case expressions are now evaluated with the constant evaluator — leniently, since the value may reference a class declared later in the file (forward references are legal; an unresolvable value degrades to a placeholder that the stub registration resolves independently in a later phase). Enum-case-valued constants are re-bound each RINIT via `php::updateConstant(..., php::getEnumCase(...))`, mirroring the array-constant mechanism, because enum case objects are request-scoped and cannot live in a MINIT zval.

Verified against Zend 8.4.13; tests + phpt included.

Part of the split of #39.

### DIFF
--- a/phpunit/code/enum-case-class-constant.php
+++ b/phpunit/code/enum-case-class-constant.php
@@ -1,0 +1,24 @@
+<?php
+enum CodegenEnum: int
+{
+    case B = 4;
+    case A = 1 + 1;
+}
+
+enum CodegenTyped
+{
+    case A;
+}
+
+class CodegenHolder
+{
+    public const CB = CodegenEnum::B;
+    public const PICKED = true ? CodegenEnum::A : CodegenEnum::B;
+    public const CodegenTyped CASE_VALUE = CodegenTyped::A;
+    public const MODE = RoundingMode::HalfEven;
+}
+
+function main(): void
+{
+    var_dump(CodegenHolder::CB === CodegenEnum::B);
+}
--- a/phpunit/src/EnumCaseAstConstantLifecycleTest.php
+++ b/phpunit/src/EnumCaseAstConstantLifecycleTest.php
@@ -1,0 +1,120 @@
+<?php
+
+use TypePhp\CompilerBase;
+use TypePhp\CompilerTest;
+
+/**
+ * Lifecycle contract for persistent enum-case AST class constants: Zend's
+ * internal-class teardown (destroy_zend_class) tolerates them only after the
+ * module's MSHUTDOWN released them. The generated module must therefore
+ * (a) refuse to start as a MODULE_TEMPORARY (dl()-loaded) module before any
+ * class is registered, because module_destructor() destroys temporary-module
+ * classes before the shutdown callback runs, and (b) order every fallible
+ * MINIT step before the first class registration, so a FAILURE return (which
+ * suppresses MSHUTDOWN) can never leave a foreign AST in the class table.
+ */
+final class EnumCaseAstConstantLifecycleTest extends \BaseTest
+{
+    public function testModuleTemporaryGuardPrecedesEveryRegistrationStep(): void
+    {
+        $minit = $this->generateMinitBody('enum-case-class-constant.php', 'ast_lifecycle_guard');
+
+        $guardPos = strpos($minit, 'if (type == MODULE_TEMPORARY) {');
+        self::assertIsInt($guardPos, 'MINIT must reject dl()-loaded temporary modules');
+        self::assertStringContainsString(
+            'registers enum-case class constants that must be released by MSHUTDOWN',
+            $minit,
+        );
+        self::assertStringContainsString('Load the extension from php.ini instead.', $minit);
+
+        $handlersPos = strpos($minit, 'typephp_install_reflection_attribute_handlers()');
+        $firstRegisterPos = strpos($minit, 'register_class_');
+        self::assertIsInt($handlersPos);
+        self::assertIsInt($firstRegisterPos);
+        self::assertLessThan($handlersPos, $guardPos, 'the lifecycle guard must be the first MINIT statement');
+        self::assertLessThan($firstRegisterPos, $guardPos, 'the lifecycle guard must precede every class registration');
+    }
+
+    public function testAstConstantRegistrationIsOrderedAfterEveryFallibleMinitStep(): void
+    {
+        $minit = $this->generateMinitBody('enum-case-class-constant.php', 'ast_lifecycle_order');
+
+        $firstRegisterPos = strpos($minit, 'register_class_');
+        $lastFailurePos = strrpos($minit, 'return FAILURE;');
+        self::assertIsInt($firstRegisterPos);
+        self::assertIsInt($lastFailurePos);
+        self::assertLessThan(
+            $firstRegisterPos,
+            $lastFailurePos,
+            'no MINIT step after the first class registration may return FAILURE: '
+                . 'a failed MINIT never reaches MSHUTDOWN, so the persistent class '
+                . 'table would keep an AST that destroy_zend_class() cannot handle',
+        );
+        self::assertGreaterThan(
+            strrpos($minit, 'register_class_'),
+            strpos($minit, 'return SUCCESS;'),
+        );
+    }
+
+    public function testMshutdownReleasesAstConstantsBeforeAnyOtherTeardown(): void
+    {
+        $extension = $this->generateExtension('enum-case-class-constant.php', 'ast_lifecycle_shutdown');
+        $mshutdown = $this->sliceFunction($extension, 'PHP_MSHUTDOWN_FUNCTION', 'THREAD_LOCAL zval globals_array');
+
+        $releasePos = strpos($mshutdown, 'typephp_release_ast_constants_enum_case_class_constant();');
+        self::assertIsInt($releasePos, 'MSHUTDOWN must release the persistent AST constants');
+        // The release must run before anything else so the class table is
+        // Zend-safe no matter what the rest of the teardown does.
+        $firstStatementPos = strpos($mshutdown, ';');
+        self::assertSame($firstStatementPos, $releasePos + strlen('typephp_release_ast_constants_enum_case_class_constant();') - 1);
+    }
+
+    public function testModulesWithoutAstConstantsCarryNeitherGuardNorRelease(): void
+    {
+        $extension = $this->generateExtension('class-constant-codegen.php', 'ast_lifecycle_none');
+
+        self::assertStringNotContainsString('MODULE_TEMPORARY', $extension);
+        self::assertStringNotContainsString('typephp_release_ast_constants_', $extension);
+    }
+
+    private function generateMinitBody(string $fixture, string $target): string
+    {
+        return $this->sliceFunction(
+            $this->generateExtension($fixture, $target),
+            'PHP_MINIT_FUNCTION',
+            'PHP_MSHUTDOWN_FUNCTION',
+        );
+    }
+
+    private function generateExtension(string $fixture, string $target): string
+    {
+        global $translator;
+
+        $compiler = CompilerTest::create(TYPEPHP_ROOT_PATH);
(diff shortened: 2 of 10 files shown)


matyhtf (changes requested):
Thank you for isolating this fix from #39. Preserving enum-case object identity for class constants is necessary, and the added direct-case tests pass. However, this implementation has lifecycle and compatibility blockers that need to be addressed before merging.

1. **The RINIT/RSHUTDOWN rebinding mutates a persistent class-constant table and is not ZTS-safe**

The generated class constant is initially a scalar placeholder, so the internal class does not get `ZEND_ACC_HAS_AST_CONSTANTS` or a request-local mutable constants table. `php::updateConstant()` therefore reaches the persistent `ce->constants_table` and replaces `zend_class_constant::value` directly with a request-owned enum object.

With concurrent ZTS requests, request A and request B can overwrite the same constant, and one request can reset it to null while the other is still using it. This can also expose an object owned by one request to another request. Mirroring the existing array-constant code does not make this safe; that helper path appears to have the same underlying limitation for refcounted request values.

Please use Zend's request-local mutable constant table, or register an appropriate persistent `IS_CONSTANT_AST` and let Zend perform per-request separation, evaluation, and cleanup. A request-owned object must never be written into the shared persistent class-entry table.

2. **Valid enum-case class constants are still unsupported**

These cases fail against the updated branch:

- Internal enum case:

```php
class InternalEnumCaseHolder
{
    public const MODE = RoundingMode::HalfEven;
}
```

Compilation aborts in `gen_stub.php` with `Invalid default value: RoundingMode Enum ... type: object`.

- Typed class constant:

```php
enum TypedCase { case A; }

class TypedHolder
{
    public const TypedCase CASE_VALUE = TypedCase::A;
}
```

Compilation aborts because the scalar placeholder `"A"` is not valid for the declared `TypedCase` type.

Both are valid PHP 8.4 declarations and should preserve case identity through static access, `constant()`, and reflection.

3. **Only direct `ClassConstFetch` chains are recognized**

`resolveEnumCaseClassConstant()` returns null for valid constant expressions that evaluate to an enum case:

```php
enum E { case A; case B; }

class K
{
    public const VALUE = true ? E::A : E::B;
}
```

The compiled direct access evaluates correctly, but `constant('K::VALUE') === E::A` is `false`, because the registered constant remains the folded string placeholder. The solution needs to represent/evaluate enum-case-valued constant expressions, not only syntactically follow direct `E::Case` and constant aliases.

4. **Constant evaluation should not be moved into prepare**

`evaluateEnumCaseValue()` eagerly evaluates the case expression while declarations are still being collected. The need to catch forward-reference failures demonstrates that the symbol environment is incomplete at this phase. It also catches every `Throwable`, which can hide genuine compiler defects, and overloads `null` to mean both a pure case and an unresolved backed case.

Please keep the value AST and explicit backed/pure state during prepare, then resolve it after declarations are complete in convert. This follows TypePHP's declaration-only prepare model and removes the broad exception-based fallback.

Please add PHPT coverage for internal enum cases, typed enum-case class constants, enum-case-valued conditional expressions, and repeated request/ZTS-safe lifecycle behavior. The current direct user-enum tests and the forward-reference path pass, but the cases above do not.

AlessioGiacobbe:
Rewritten from scratch along the lines you suggested — the branch is now a single commit implementing the engine-managed design, rebased on latest master.

1. **No more RINIT/RSHUTDOWN rebinding.** The constant is registered at MINIT as a persistent `IS_CONSTANT_AST` holding the `Enum::Case` fetch — the same mechanism internal enums use for their own case constants. Zend separates the class constants table into request-local mutable storage on first access, evaluates the fetch there, and cleans it up at request shutdown: no request-owned object ever touches the persistent class entry, safe under concurrent ZTS requests, and lazy evaluation also removes any registration-order sensitivity. I validated the mechanism against PHP 8.4.13 with a standalone test extension before wiring it into gen_stub (static access, `constant()`, and reflection all preserve identity; typed constants keep their declared type).

2. **Internal enum cases and typed constants now work.** `const MODE = RoundingMode::HalfEven;` registers the same AST (no more gen_stub abort), and `const TypedCase CASE_VALUE = TypedCase::A;` goes through `zend_declare_typed_class_constant` with the AST value — no scalar placeholder exists anymore, so there is nothing to violate the declared type.

3. **Constant expressions preserve identity.** Case identity flows through compile-time constant evaluation as an `EnumCaseRef` value instead of a scalar, so `true ? E::A : E::B` folds to the `E::A` identity and `constant('K::VALUE') === E::A` is true; constant chains (`const REF = K::CB;`) follow for free. The runtime expression path (`php::getEnumCase`) is untouched.

4. **No more eager evaluation in prepare.** The `withThrowingDiagnostics` machinery and the catch-all are gone. The preprocessor only records literal backing values (fixing the original `->value`-off-an-expression warning); nothing at compile time needs the evaluated scalar of an expression-valued case — gen_stub evaluates the registration value from the AST itself, and `null` again means exactly "pure case".

The PHPT now covers all your cases with Zend-8.4-validated expectations: internal enum case, typed constant, ternary expression, constant chain, `constant()`, dynamic class access, reflection (value and type), and the computed backing value of `case A = 1 + 1;`.

AlessioGiacobbe:
CI failures fixed (new commit):

- **Property/parameter defaults** — the AST representation had leaked into their persistent tables, which reject refcounted zvals ("Internal zvals cannot be refcounted" at startup). `EvaluatedValue` now carries the case identity in a dedicated field while its `value` degrades to exactly what those consumers read before (host case object for internal enums, literal backing/case name for compiled ones), and only class-constant registration opts into the AST — `imported-enum-defaults` and `default-initialization-paths` are back on their unchanged runtime-restore path.
- **`K::VALUE->value`** — a C++ precedence bug my own PHPT caught: the folded ternary of `const VALUE = cond ? E::A : E::B;` bound `.attr("value")` to its else branch only. Folded constant operands with top-level operators are now parenthesized before a member access is appended.

Verified locally: property defaults emit the legacy scalar again (zero ASTs outside constants), the five constant registrations still emit the AST, the full 1138-file phpt corpus front-end sweeps clean, and the related PHPUnit suites show only the pre-existing environmental failures.

matyhtf (changes requested):
This still cannot be merged because the persistent AST representation violates Zend's internal-class teardown contract.

Running the PR's own `enum-case-class-constant.phpt` with a PHP 8.4 debug build produces all expected output, then aborts during shutdown:

```text
zend_opcode.c:499: destroy_zend_class:
Assertion `Z_ASTVAL(c->value)->kind == ZEND_AST_CONST_ENUM_INIT' failed.
Aborted
```

`initializeEnumCaseZval()` registers an internal class constant whose persistent value is `IS_CONSTANT_AST` with root kind `ZEND_AST_CLASS_CONST`. In both PHP 8.4 and 8.5, `destroy_zend_class()` explicitly asserts that persistent constant ASTs owned by internal classes are `ZEND_AST_CONST_ENUM_INIT`. Release PHP builds used by CI compile out this assertion, which is why all CI jobs pass.

There is also an ownership leak: the two `zend_ast_zval` children are allocated by separate `pemalloc()` calls, while Zend's internal-class teardown only frees the allocation referenced by `Z_AST(c->value)`. Zend's own persistent enum AST builder stores the ref, root, and children in one contiguous allocation.

Changing the root kind to `ZEND_AST_CONST_ENUM_INIT` is not sufficient: that node constructs another enum object instead of fetching the canonical registered case, so case identity would be broken again.

Please redesign this around a representation with a supported and complete lifecycle, or add explicit ownership/teardown infrastructure for the custom persistent AST. The PHPT must pass through process shutdown under PHP 8.4 and 8.5 debug builds, not only release CI.

AlessioGiacobbe:
Lifecycle redesigned per the review (rebased on current master):

- **One contiguous persistent allocation** now holds the ast_ref, the CLASS_CONST root, and both zval children — mirroring Zend's own persistent enum AST builder — so ownership is a single block and the children can no longer leak.
- **Explicit teardown before class destruction**: every generated file with AST constants emits a `typephp_release_ast_constants_*` function that the module's MSHUTDOWN calls before Zend's class teardown runs. It frees the block and restores the constant slot to null, so `destroy_zend_class()` never encounters a non-`CONST_ENUM_INIT` persistent AST — the debug assertion is unreachable by construction, on 8.4 and 8.5 alike. Request-local mutable copies are untouched (no request is live at MSHUTDOWN).
- `CONST_ENUM_INIT` itself was ruled out for the reason you gave: it constructs a new case object instead of fetching the canonical one, which would break identity again.

I validated the mechanism with a standalone test extension against PHP 8.4.13 (contiguous layout, identity preserved through static access/`constant()`/reflection, clean MSHUTDOWN release and process exit). I don't have a debug build locally, so the assertion itself I can only make unreachable by construction — the constant is no longer `IS_CONSTANT_AST` by the time class teardown runs; if you can re-run the PHPT on your debug build, that would be the definitive check.

matyhtf (changes requested):

There is still a lifecycle blocker in the AST-constant teardown design. The generated `typephp_release_ast_constants_*()` function is called only from MSHUTDOWN. That ordering is valid for a normal persistent module shutdown, but it is not a general pre-class-destruction hook:

- For `MODULE_TEMPORARY` modules loaded through `dl()`, Zend `module_destructor()` calls `clean_module_classes()` before the module shutdown callback. The foreign `ZEND_AST_CLASS_CONST` therefore reaches `destroy_zend_class()` first and still triggers the debug assertion.
- If MINIT registers one of these constants and then fails before `module_started` becomes true, MSHUTDOWN is not guaranteed to run either.

Please redesign this so the persistent class table never depends on a later MSHUTDOWN rewrite for Zend-safe destruction, or explicitly reject every lifecycle that cannot provide that guarantee before registering any affected class. Add a PHP debug/ZTS regression that exercises the supported shutdown and module-load paths; checking expected runtime output alone will not catch the shutdown assertion.

AlessioGiacobbe:
Rebased onto current master. The lifecycle contract is now enforced instead of assumed:

- When the module declares any enum-case AST constants, the generated MINIT opens with a `MODULE_TEMPORARY` guard (`zend_error` + `return FAILURE`) **before any class registration** — a `dl()` load is rejected with an empty class table, so `clean_module_classes()` never sees a foreign `ZEND_AST_CLASS_CONST`.
- MINIT is restructured so every step that can return FAILURE precedes the first `register_class_*()` call; the AST constants are installed by the infallible registration tail, so a failing MINIT can never leave a foreign AST in the persistent tables. The generator itself throws at build time if a future change emits a FAILURE return after registration begins.
- The MSHUTDOWN release is unchanged and remains the supported persistent-module path.

Tests: `EnumCaseAstConstantLifecycleTest` asserts the guard exists exactly when AST constants exist, precedes every registration step, that no FAILURE return follows the first registration, and that the release call is MSHUTDOWN's first statement. The phpt now also carries a never-accessed constant so a pristine persistent AST goes through full process shutdown (a debug build turns a violation into an output mismatch via the assert on stderr). A runnable `dl()`-path phpt is not feasible in this harness — tests are linked into standalone binaries with persistently registered modules only — which is documented in the test; that lifecycle is now rejected at MINIT rather than left undefined.

AlessioGiacobbe:
The single PHPT failure on the previous run was `tests/compiler/string_method/method.phpt`, unrelated to this PR: line 79 asserts `Assert::notEq($text->shuffle(), str_shuffle($text))` — two independent shuffles of a 6-character string, which collide with probability 1/720 per run. Re-pushed to re-run CI; the assertion itself might be worth pinning against the original string instead of a second shuffle.

PR #52, opened 2026-09-01
fix(codegen): keep Zend operand read order around hoisted side effects

AlessioGiacobbe:
When a later call argument or concat operand hoisted a side effect, it executed before earlier operands were read: `$j = 1; two($j, $j = 5)` printed `"5,5"` where Zend SENDs left-to-right and prints `"1,5"`; `$m . "," . ($m = 9)` printed `"9,9"` instead of `"1,9"`.

Earlier plain-variable operands are now snapshotted at their exact Zend read positions when a later operand materializes statements. Concat read positions were probed against Zend empirically (the VM reads the first two chain items together after both items' side effects), and plain arithmetic (`$k + ($k = 5)`) is deliberately unchanged — Zend itself reads the CV at the op, so the existing codegen already matches; a no-regression test pins that.

Verified against Zend 8.4.13; codegen tests + phpt included.

Part of the split of #39.

### DIFF
--- a/phpunit/code/eval-order-side-effects.php
+++ b/phpunit/code/eval-order-side-effects.php
@@ -1,0 +1,24 @@
+<?php
+
+function pair(int $a, int $b): string
+{
+    return $a . ',' . $b;
+}
+
+function callArgOrder(): string
+{
+    $j = 1;
+    return pair($j, $j = 5);
+}
+
+function concatOrder(): string
+{
+    $m = 1;
+    return $m . ',' . ($m = 9);
+}
+
+function plainArithmeticUnchanged(): int
+{
+    $k = 1;
+    return $k + ($k = 5);
+}
--- a/phpunit/code/eval-order-side-effects.php
+++ b/phpunit/code/eval-order-side-effects.php
@@ -22,3 +22,20 @@ function plainArithmeticUnchanged(): int
     $k = 1;
     return $k + ($k = 5);
 }
+
+function pairValue(mixed $a, mixed $b): string
+{
+    return $a . ',' . $b;
+}
+
+function castWrappedCallArgOrder(): string
+{
+    $i = 1;
+    return pair($i, (int) ($i = 5));
+}
+
+function notWrappedCallArgOrder(): string
+{
+    $k = 1;
+    return pairValue($k, !($k = 0));
+}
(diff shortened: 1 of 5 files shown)


matyhtf (changes requested):
The current operand classifier misses side effects wrapped in non-binary expressions, so the argument-order bug is still observable.

For example:

```php
function pairValue(mixed $a, mixed $b): string
{
    return $a . ',' . $b;
}

$i = 1;
var_dump(pairValue($i, (int) ($i = 5)));

$k = 1;
var_dump(pairValue($k, !($k = 0)));
```

Zend PHP outputs:

```text
string(3) "1,5"
string(3) "1,1"
```

This branch outputs:

```text
string(3) "5,5"
string(3) "0,1"
```

`shouldMaterializeOrderedOperand()` only descends through `Expr\BinaryOp`. An assignment wrapped by `Expr\Cast` or `BooleanNot` is therefore classified as non-hoisting, even though lowering it appends captured statements. The earlier by-value argument is not snapshotted and the nested assignment overtakes its read.

Please make the side-effect/hoisting detection structurally recursive through expression wrappers such as casts, unary expressions, boolean-not, bitwise-not, and error suppression, or derive the decision from the captured statements produced by lowering. A generic AST walk must stop at nested Closure/ArrowFunction bodies because their bodies are not evaluated at creation time.

Add PHPT coverage for at least the cast and boolean-not examples above. The existing tests pass but do not cover wrapped side effects.

AlessioGiacobbe:
Fixed and rebased on current master. `shouldMaterializeOrderedOperand()` is now a structural recursion: side-effecting leaf kinds (calls, assignments, inc/dec, new, yield/throw/eval/shell-exec, …) classify directly, `Closure`/`ArrowFunction` stop the walk (their bodies don't run at creation), and every other expression recurses through all `Expr` sub-nodes — so casts, unary operators, boolean/bitwise not, `@`, interpolation, and any future wrapper are covered without drift.

On your suggested alternative (deriving from captured statements): I investigated it and it can't work here — the classifier runs *before* lowering (earlier arguments must be snapshotted before the later operand is emitted), and lowering doesn't always hoist: `(int)($i = 5)` lowers to the inline expression `php::toInt(i = 5LL)` with zero captured statements, which is exactly the reported bug. The rationale is in the commit body.

Both your repros now match Zend (`"1,5"` and `"1,1"`); the PHPT gains eight Zend-validated expectations (cast, boolean-not, unary-minus, `@`, and cast/not inside concat), and `$k + ($k = 5)` codegen stays byte-identical (no-regression test still green).

PR #54, opened 2026-09-01
fix(translator): let a trailing child variadic absorb parent parameters

AlessioGiacobbe:
A trailing child variadic absorbing parent parameters was rejected: parent `f(int $a, int $b)`, child `f(int ...$args)` failed "must be compatible", breaking the common decorator/proxy pattern (Zend accepts it). The per-position loop required exact variadic equality at each index.

The loop now follows Zend's rules: a variadic parent requires a variadic child; a trailing child variadic covers every remaining parent position (contravariant type, matching by-ref-ness per position); extra child parameters must stay optional. The pre-existing testVariadicMismatch encoded the rejects-valid behavior (verified against Zend) and was updated to assert compilation.

Verified against Zend 8.4.13, 11-case probe matrix.

Part of the split of #39.

### DIFF
--- a/phpunit/code/override_parent_variadic_child_extra.php
+++ b/phpunit/code/override_parent_variadic_child_extra.php
@@ -1,0 +1,12 @@
+<?php
+class A
+{
+    public function f(int ...$args): void {}
+}
+
+class B extends A
+{
+    public function f(int $a = 1, int ...$rest): void {}
+}
+
+function main() {}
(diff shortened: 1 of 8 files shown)


matyhtf (approved):
The implementation matches Zend's variadic override rules, including repeated comparison against the parent/child variadic slot, contravariant parameter checks, invariant by-reference mode, and required/optional arity constraints. The focused and existing inheritance suites pass locally (75 tests). No blocking issue found.

PR #55, opened 2026-09-01
fix(translator): validate overrides of built-in class methods

AlessioGiacobbe:
Overrides of methods inherited from Zend built-in classes were never signature-checked — only PRIVATE/FINAL flags were consulted: `class C extends ArrayObject { public function offsetGet(int $key): string {...} }` and `public static function count(): int` both compiled (Zend: fatal). Compiled classes register as internal classes, where ZendVM does not re-run these checks, so the invalid vtables reached runtime.

A reflection-based validator now enforces the Zend ruleset against built-in parents: visibility may widen but not narrow, staticness must match, a by-ref return may be added but not dropped, required-argument counts, per-position by-ref-ness with variadic absorption, parameter contravariance, and return covariance — mapping ReflectionType (named/nullable/union/intersection, self/parent/static) into the existing DNF comparison machinery. TENTATIVE return types are exempt, matching Zend, which only deprecates a tentative mismatch (`ArrayObject::count(): string` stays accepted).

Verified against Zend 8.4.13, 13-case probe matrix.

Part of the split of #39.

### DIFF
--- a/phpunit/code/internal_override_param_narrowed.php
+++ b/phpunit/code/internal_override_param_narrowed.php
@@ -1,0 +1,10 @@
+<?php
+class C extends ArrayObject
+{
+    public function offsetGet(int $key): string
+    {
+        return 'x';
+    }
+}
+
+function main() {}
--- a/phpunit/code/internal_override_real_return_mismatch.php
+++ b/phpunit/code/internal_override_real_return_mismatch.php
@@ -1,0 +1,10 @@
+<?php
+class C extends DateTime
+{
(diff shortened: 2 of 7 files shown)


matyhtf (approved):
The reflection-based check covers the missing fatal compatibility rules for internal parents: visibility/staticness, required arity, variadic absorption, invariant reference mode, parameter contravariance, real return covariance, and tentative-return exemption. The focused suite passes locally. No blocking issue found.

PR #56, opened 2026-09-01
fix(translator): constructor, private-method and abstract redeclaration override rules

AlessioGiacobbe:
Three gaps in the parent-method override checks, all in checkParentMethodCanBeOverridden:

- `__construct` overrides skipped every check via an early return — including `final` parent constructors (Zend: "Cannot override final method A::__construct()") and abstract-constructor signature contracts. Ordinary concrete parent constructors remain exempt from LSP variance, as in Zend (even visibility narrowing is legal there — probed).
- Redeclaring a parent's *private* method was rejected, but private methods are not inherited in PHP: any signature is legal. Verified safe for this compiler's dispatch before changing: private-resolved calls devirtualize to the declaring class's body (canDevirtualize), matching PHP's private-scope binding, and Native classes give private methods no virtual slot. Two pre-existing tests encoded the rejects-valid behavior (both run under Zend) and were updated.
- Abstract method redeclarations were never validated: turning a concrete inherited method abstract compiled (Zend: "Cannot make non abstract method A::f() abstract in class B") and abstract-over-abstract redeclarations skipped the signature check. Trait-originated abstract requirements stay exempt — Zend lets an inherited concrete method satisfy them (probed).

Verified against Zend 8.4.13 for every rule and direction.

Part of the split of #39.

### DIFF
--- a/phpunit/code/abstract_redeclare_concrete.php
+++ b/phpunit/code/abstract_redeclare_concrete.php
@@ -1,0 +1,5 @@
+<?php
+class A { public function f(): int { return 1; } }
+abstract class B extends A { abstract public function f(): int; }
+
+function main() {}
--- a/phpunit/code/abstract_redeclare_incompatible.php
+++ b/phpunit/code/abstract_redeclare_incompatible.php
@@ -1,0 +1,5 @@
+<?php
+abstract class A { abstract public function f(): int; }
+abstract class B extends A { abstract public function f(): string; }
+
+function main() {}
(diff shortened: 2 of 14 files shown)


matyhtf (approved):
The constructor, private-method redeclaration, and abstract-redeclaration rules match Zend behavior. I also ran an end-to-end private dispatch case: a parent-scope call still reaches the parent's private implementation while child/external calls reach the child declaration. Focused tests pass. Non-blocking: the checkParentMethodCanBeOverridden() docblock still says private methods cannot be overridden and should be updated when this is merged.

matyhtf:
@copilot resolve the merge conflicts in this pull request

PR #57, opened 2026-09-01
fix(translator): reject static/instance property redeclaration mismatch

AlessioGiacobbe:
Redeclaring a static property as non-static (or the reverse) compiled: `class A { public static int $x = 1; } class B extends A { public int $x = 2; }` — Zend fatals "Cannot redeclare static A::$x as non static B::$x". checkPropertyOverride compared type, visibility, readonly, and final, but never the STATIC flag.

Both directions now rejected with Zend's message.

Part of the split of #39.

### DIFF
--- a/phpunit/code/property_nonstatic_mismatch.php
+++ b/phpunit/code/property_nonstatic_mismatch.php
@@ -1,0 +1,12 @@
+<?php
+class A
+{
+    public int $x = 1;
+}
+
+class B extends A
+{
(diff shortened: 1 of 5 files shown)


matyhtf (approved):
The static/instance storage-kind check is correctly placed in the existing property override path, rejects both mismatch directions with stable diagnostics, and leaves matching redeclarations unchanged. No blocking issue found.

PR #58, opened 2026-09-01
fix(translator): covariant typed class constants and interface constant contracts

AlessioGiacobbe:
Two related gaps in class-constant type contracts:

- Typed class constants were forced invariant: `const int X` overriding `const int|string X` was rejected, while PHP 8.3 constants are covariant (narrowing legal, widening not — widening stays rejected). ConstantDef now records the declared type's accepted-types DNF (composite types were previously collapsed to one variant type) and the override check uses the covariance machinery.
- Interface constants were never validated at all: overriding a `final const`, incompatible types, non-public visibility, and the same constant name arriving ambiguously from two interfaces (including via parent classes and enum implements) all compiled. An effective-constants table preserving the original declaring interface now enforces Zend's rules for classes, enums, and interface-extends-interface.

Verified against Zend 8.4.13, 32-case probe matrix.

Part of the split of #39.

### DIFF
--- a/phpunit/code/const_override_covariant.php
+++ b/phpunit/code/const_override_covariant.php
@@ -1,0 +1,17 @@
+<?php
+class A
+{
+    const int|string X = 1;
+    const mixed M = 1;
+    const ?int N = null;
+}
+
+class B extends A
+{
+    // PHP 8.3 typed constants are covariant: narrowing is allowed.
+    const int X = 2;
+    const string M = 'a';
+    const int N = 5;
+}
+
+function main() {}
--- a/phpunit/code/const_override_unrelated.php
+++ b/phpunit/code/const_override_unrelated.php
@@ -1,0 +1,12 @@
+<?php
+class A
+{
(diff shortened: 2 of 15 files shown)


matyhtf (approved):
The constant covariance and effective-interface-constant table logic are consistent with PHP 8.4 behavior. I also checked direct/transitive final constants, ambiguity ordering, inherited class constants, trait-injected constants, typed narrowing, and integration with the current master; the focused tests pass and the branch merges cleanly.

Non-blocking follow-up: these new tests only exercise the compiler's dry generation path. A real native build of `const_override_covariant.php` currently fails because the existing `gen_stub.php` emits `ZVAL_LONG(&const_N_value, );` for `const ?int N = null`; DNF/intersection class-constant types also hit the existing `PropertyInfo` assertion in `VariableLike::getTypeCode()`. Those are generator gaps outside this PR's modified contract-validation path and should be covered by end-to-end PHPTs separately.

PR #59, opened 2026-09-01
fix: enforce interface declaration and merge rules

AlessioGiacobbe:
Interface declaration and merge rules were unenforced:

- Interface methods with bodies compiled (the body silently became dead code); `final`, `private`/`protected`, and explicit `abstract` modifiers were accepted; non-public interface constants and explicitly-abstract hooked properties too. All are Zend compile fatals (each probed for the exact rule and message).
- `interface I extends SomeClass` either fataled with a misleading missing-symbol message or compiled when the class was declared later; both cases now fail with Zend's "cannot implement … it is not an interface".
- Same-name methods arriving from several extended interfaces — or from several interfaces a class implements without defining the method — were never cross-checked: `interface J extends I1, I2` compiled with mutually incompatible `f()` declarations. The first-seen declaration is now validated as an override of every later one, matching Zend's merge order; diamond inheritance of one original declaration never conflicts, and a method the class chain defines silences the pairwise check (Zend-probed: a `never` return satisfying both incompatible declarations compiles).

Part of the split of #39.

### DIFF
--- a/phpunit/code/interface_collision_unimplemented.php
+++ b/phpunit/code/interface_collision_unimplemented.php
@@ -1,0 +1,6 @@
+<?php
+interface I1 { public function f(): int; }
+interface I2 { public function f(): string; }
+abstract class C implements I1, I2 {}
+
+function main() {}
--- a/phpunit/code/interface_collision_valid.php
+++ b/phpunit/code/interface_collision_valid.php
@@ -1,0 +1,17 @@
+<?php
+// Compatible multi-extends (covariant), diamond inheritance, and a class
+// method that satisfies both incompatible declarations are all legal.
+interface I1 { public function f(): int; }
+interface I2 { public function f(): int|string; }
+interface J extends I1, I2 {}
+
+interface Base { public function g(): int; }
+interface B1 extends Base {}
+interface B2 extends Base {}
+interface Diamond extends B1, B2 {}
+
+interface S1 { public function h(): int; }
+interface S2 { public function h(): string; }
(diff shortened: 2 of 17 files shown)


matyhtf (changes requested):
`getEffectiveInterfaceMethodTable()` recursively memoizes only after the parent table is complete, but it has no in-progress/cycle guard. A cyclic interface graph therefore recurses indefinitely instead of producing a compiler diagnostic:

```php
interface A extends B {}
interface B extends A {}
function main(): void {}
```

Please add a `visiting` set (cleared with `finally`) or validate the inheritance graph before recursive table construction, and add a negative test proving this input fails promptly with a stable fatal diagnostic.

PR #58 now adds cycle protection to the analogous effective constant table. Please rebase on current `master`, but do not rely only on the constant-validation call order: this method-table helper should remain safe independently.

AlessioGiacobbe:
Fixed and rebased on current master (the rebase adopted master's constants machinery from #58 verbatim, including its new cycle guard, and re-layered only this PR's method-table work).

`getEffectiveInterfaceMethodTable()` now mirrors that guard exactly: a `visiting` set cleared with `try/finally`, and a stable fatal — ``Interface inheritance cycle detected at `X` `` — the same diagnostic shape master's constant-table helper emits. The helper is safe independently of any validation call order: the new test drives it directly (bypassing every other pass) and it fails promptly instead of recursing; end-to-end compilation of `interface A extends B {} interface B extends A {}` fails with the same message. Diamond graphs still compile (covered by the existing collision-valid fixture).

13/13 focused tests green; full suite and sweep failure sets byte-identical to base.

PR #60, opened 2026-09-01
fix(translator): trait member value comparison and adaptation validation

AlessioGiacobbe:
Two trait-composition gaps:

- Trait constant/property conflicts were compared by pretty-printed source text: two traits declaring `const int X = 1 + 1;` and `const int X = 2;` (or defaults spelled `[1, 2]` vs `array(1, 2)`) were rejected as conflicting, while Zend compares evaluated values. Values are now compared by evaluation — with Zend's declaration-time int→float coercion for float-typed members — falling back to source-text equality only when a value cannot be evaluated at compile time. Different values, visibility, or types still conflict.
- Trait adaptations referencing nonexistent methods were silently ignored: `use A { missing as g; }` and `use A, B { B::f insteadof A; }` (with no `B::f`) both compiled — a typo in an adaptation did nothing. Every alias and precedence rule is now verified against the composed methods, with Zend's diagnostics; an unqualified alias is registered under every used trait, so its variants share one group satisfied by any match, and precedence rules also verify the named traits are actually used.

Verified against Zend 8.4.13; nested-trait aliasing covered.

Part of the split of #39.

### DIFF
--- a/phpunit/code/trait_adaptations_valid.php
+++ b/phpunit/code/trait_adaptations_valid.php
@@ -1,0 +1,46 @@
+<?php
+trait Inner
+{
+    public function f(): void {}
+}
+
+trait Outer
+{
+    use Inner;
+}
+
+// Aliases resolve methods arriving from NESTED traits, both unqualified and
+// qualified with the directly-used trait.
+class UsesNested
+{
+    use Outer {
+        f as g;
+    }
+}
+
+class UsesNestedQualified
+{
+    use Outer {
+        Outer::f as h;
+    }
+}
+
+trait Winner
+{
+    public function m(): void {}
+}
+
+trait Loser
+{
+    public function other(): void {}
+}
+
+// The overridden trait need not declare the method a precedence rule names.
+class PrecedenceLoserMissing
+{
+    use Winner, Loser {
+        Winner::m insteadof Loser;
+    }
+}
+
+function main() {}
--- a/phpunit/code/trait_alias_missing_method.php
+++ b/phpunit/code/trait_alias_missing_method.php
@@ -1,0 +1,14 @@
+<?php
+trait A
+{
+    public function f(): void {}
+}
+
+class C
+{
+    use A {
+        missing as g;
+    }
+}
+
+function main() {}
--- a/phpunit/code/trait_alias_missing_qualified.php
+++ b/phpunit/code/trait_alias_missing_qualified.php
@@ -1,0 +1,14 @@
+<?php
+trait A
+{
+    public function f(): void {}
+}
+
+class C
+{
+    use A {
+        A::missing as g;
+    }
+}
+
+function main() {}
--- a/phpunit/code/trait_alias_trait_not_used.php
+++ b/phpunit/code/trait_alias_trait_not_used.php
@@ -1,0 +1,19 @@
+<?php
+trait A
+{
+    public function f(): void {}
+}
+
+trait B
+{
+    public function h(): void {}
+}
+
+class C
+{
+    use A {
+        B::h as g;
+    }
+}
+
+function main() {}
--- a/phpunit/code/trait_const_enum_backed_conflict.php
+++ b/phpunit/code/trait_const_enum_backed_conflict.php
@@ -1,0 +1,30 @@
+<?php
+// A shared backing scalar does not make cases of two different backed enums
+// the same value: Zend compares the case objects, not their backing values.
+enum E1: string
+{
+    case Value = 'x';
+}
+
+enum E2: string
+{
+    case Value = 'x';
+}
+
+trait T1
+{
+    const X = E1::Value;
+}
+
+trait T2
+{
+    const X = E2::Value;
+}
+
+class C
+{
+    use T1;
+    use T2;
+}
+
+function main() {}
--- a/phpunit/code/trait_const_enum_case_array_conflict.php
+++ b/phpunit/code/trait_const_enum_case_array_conflict.php
@@ -1,0 +1,30 @@
+<?php
+// Enum-case identity nests recursively: arrays holding same-named cases of
+// two DIFFERENT enums hold distinct case objects, so the definitions conflict.
+enum E1
+{
+    case Value;
+}
+
+enum E2
+{
+    case Value;
+}
+
+trait T1
+{
+    const X = [E1::Value];
+}
+
+trait T2
+{
+    const X = [E2::Value];
+}
+
+class C
+{
+    use T1;
+    use T2;
+}
+
+function main() {}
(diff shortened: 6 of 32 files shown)


matyhtf (changes requested):
Two invalid programs still compile.

1. `traitIgnored` stores a single rule under the loser `Trait::method` key. A later rule for the same loser overwrites the earlier rule, so not every preferred method is validated:

```php
trait A { public function other(): void {} }
trait B { public function f(): void {} }
trait C { public function f(): void {} }
class X {
    use A, B, C {
        A::f insteadof B;
        C::f insteadof B;
    }
}
```

Zend reports `A precedence rule was defined for A::f but this method does not exist`; this PR compiles it. Preserve every precedence-rule record separately from (or as a list behind) the ignored-method lookup, and validate every record.

2. Trait constant identity short-circuits on `ConstantDef::$value`. Different enum classes with the same case name are both normalized to the same string and are incorrectly accepted:

```php
enum E1 { case Value; }
enum E2 { case Value; }
trait T1 { const X = E1::Value; }
trait T2 { const X = E2::Value; }
class X { use T1, T2; }
```

Zend rejects these distinct case objects; this PR compiles them. Do not use the normalized case-name string as an identity proof. Retain/compare the enum class and case identity, and add positive coverage for the same enum case plus negative coverage for different enum classes.

AlessioGiacobbe:
Both fixed, rebased on current master:

1. **Precedence rules** — every rule record is now preserved as a list behind the loser key (the `isset()`-based ignore lookup is unchanged), `prepareTraitUse` merges per-key instead of dropping earlier `use` clauses' rules, and validation iterates every record: your repro fails with Zend's exact wording (`A precedence rule was defined for `A::f` but this method does not exist`). Keeping all records also let me implement Zend's duplicate-exclusion rule ("Method of trait B was defined to be excluded multiple times" — probed), reported after the existence checks in Zend's order. One winner excluding several losers stays legal (probed).
2. **Enum case identity** — the trait-member value comparison no longer collapses cases to a scalar: the evaluation path returns a collision-proof `(enum class)::(case)` identity used only by the trait comparator (general constant folding untouched), and the premature `value === value` short-circuit — which treated every non-literal initializer as equal because they all share an empty compiled value — now defers to the evaluated comparison. All six probed combinations match Zend: same enum + same case compiles; different enums, same enum different cases, different backed enums with equal backing, and array-wrapped cases all reject.

17/17 focused tests green; sweep and full suite identical to base. One scope note: a case reached indirectly through *another class constant* still collapses to its scalar in this comparator — happy to cover that in a follow-up if you'd like it here.

matyhtf (changes requested):

The enum-case identity marker is still representable by user code. PHP strings are binary-safe, so the NUL prefix does not make this value collision-free:

```php
enum E { case A; }
trait T1 { const X = E::A; }
trait T2 { const X = "\0enum-case\0e::A"; }
class C { use T1, T2; }
```

Zend rejects the composition because an enum-case object and a string are different values. This branch evaluates both definitions to the same marker string and may accept them.

Please use an internal typed identity value that PHP constant expressions cannot construct, with stable/interned identity for repeated references to the same enum case, and add the collision above as a negative regression test. Arrays containing enum cases must preserve the same distinction recursively.

AlessioGiacobbe:
Rebased onto current master. The identity marker is no longer a string:

- Enum cases now evaluate (under identity semantics) to an interned compiler-internal object — `EnumCaseIdentity::intern(enumClass, caseName)`, one singleton per (lowercased FQ enum, case-sensitive case name) pair. `ConstExprEvaluator` can only produce scalars, arrays and null from user constant expressions, so no user-spellable value can ever be `===` to an instance; interning makes `===` a stable identity for repeated references, and PHP's `===` on arrays compares elements with `===` again, so the distinction nests recursively.
- `ENUM_CASE_IDENTITY_PREFIX` and the NUL-marker code are gone.

Tests added (each first verified against Zend 8.4.13): the exact collision program from the review (`E::A` vs `"\0enum-case\0e::A"`) now conflicts; `[E::A, 1]` in both traits composes; `[E1::Value]` vs `[E2::Value]` conflicts.

matyhtf (changes requested):
The direct enum-case comparison is now protected by an internal identity object, but indirect class constants still lose that identity.

`getClassConstValue(..., $enumCasesAsIdentity)` honors the flag for a direct enum case, while `findNativeClassConst()` calls `evaluateClassConstValue()` without forwarding it, and `resolveInheritedClassConst()` has no corresponding parameter. As a result, enum cases reached through nested or inherited class constants collapse to their scalar/case-name representation.

This currently compiles even though Zend rejects it:

```php
enum E1 { case Value; }
enum E2 { case Value; }

class H1 { public const ITEM = E1::Value; }
class H2 { public const ITEM = E2::Value; }

trait T1 { public const VALUE = H1::ITEM; }
trait T2 { public const VALUE = H2::ITEM; }

class C {
    use T1, T2;
}
```

Please propagate the enum-identity mode through `evaluateClassConstValue()`, `resolveInheritedClassConst()`, native constant lookup, and every nested/inherited constant-evaluation path. Add a positive test for the same indirect enum case and a negative test for different indirect enum cases.

AlessioGiacobbe:
Rebased onto current master and identity mode is now threaded through every constant-evaluation path: `getClassConstValue()` forwards it to the nativeConst evaluation and to `resolveInheritedClassConst()` (new parameter), the `constant()` fallbacks (linked-class refs, internal classes, global constants inside expressions) intern a live `UnitEnum` under the flag, and the ConstExprEvaluator callback already propagated it.

One note: after the rebase onto the merged enum-case class-constant work, your program no longer compiled — but for the wrong reason: indirect constants evaluated to fresh `EnumCaseRef` objects, and `===` on distinct objects is always false, so the *same* case reached indirectly was also being rejected while Zend composes it. Threading identity mode fixes both directions.

Tests (all verified against Zend 8.4.13 first): your exact program (negative), same case via `H1::ITEM` in both traits (positive), same/different cases via constants inherited from a parent class — covering `resolveInheritedClassConst` — plus mixed direct/indirect and internal-enum spot checks. Non-identity behavior (the `EnumCaseRef` general path, `enumCaseLegacyValue()`) is untouched.

matyhtf:
Thank you for identifying these trait-composition issues and for the detailed follow-up work.\n\nWe are going to close this PR because the remaining constant-evaluation changes overlap with other compiler work and need to be completed in a coordinated way on our side. We will fix the related problems directly in master and will reuse or adapt the valuable test cases and relevant implementation ideas from this PR.

matyhtf:
Additional context on the closure decision: this work reaches beyond isolated trait checks. Trait declaration equivalence, constant-expression evaluation, enum-case identity, and the boundary between prepare, compose, and convert phases are core parts of the TypePHP language architecture.\n\nWe need to reconsider these rules together at the language-design level before choosing an implementation. For that reason, this PR will remain closed while we redesign and implement the solution internally. We will retain the reported cases, tests, and useful implementation ideas as references.

PR #61, opened 2026-09-01
fix(preprocessor): enforce Zend enum declaration rules

AlessioGiacobbe:
Enum declaration rules were entirely unenforced — all of these compiled (each a Zend compile fatal, probed for the exact rule and message): properties in enums; the forbidden magic-method set (__construct, __destruct, __clone, __get/__set/__unset/__isset, __sleep/__wakeup, __set_state, __serialize/__unserialize, __toString, __debugInfo — the full set probed one by one; __call/__callStatic/__invoke stay legal); a case with a value in a non-backed enum and a case without one in a backed enum; duplicate case names and case/const clashes; backing types other than int|string; explicit `implements UnitEnum`/`BackedEnum`; abstract methods in enum bodies.

Enum ClassDef flags now also carry FINAL (mirroring ZEND_ACC_FINAL on enum class entries), so `class B extends SomeEnum` is rejected by the existing final-class check with Zend's own wording, and final-class devirtualization legitimately applies to enums.

Deliberately not added, because Zend evaluates them lazily at runtime: case-value/backing-type mismatches and duplicate case values.

Part of the split of #39.

### DIFF
--- a/phpunit/code/enum_case_cross_file_def.php
+++ b/phpunit/code/enum_case_cross_file_def.php
@@ -1,0 +1,20 @@
+<?php
+
+// Declaring half of the cross-file pair (see enum_case_cross_file_ref.php):
+// the case expression names Provider through this file's `use` import, so a
+// lazy evaluation triggered from another file must restore this context.
+namespace Lib {
+    class Provider
+    {
+        public const BASE = 40;
+    }
+}
+
+namespace A {
+    use Lib\Provider;
+
+    enum E: int
+    {
+        case X = Provider::BASE + 2;
+    }
+}
--- a/phpunit/code/enum_case_cross_file_ref.php
+++ b/phpunit/code/enum_case_cross_file_ref.php
@@ -1,0 +1,19 @@
+<?php
+
+// Referencing half of the cross-file pair (see enum_case_cross_file_def.php).
+// This file is converted first, so Holder's constant initializer performs the
+// first fetch of A\E::X while namespace Consumer is active.
+namespace Consumer {
+    class Holder
+    {
+        public const REF = \A\E::X;
+    }
+}
+
+namespace {
+    function main()
+    {
+        var_dump(\Consumer\Holder::REF);
+        var_dump(\A\E::X->value);
+    }
+}
--- a/phpunit/code/enum_case_cross_namespace.php
+++ b/phpunit/code/enum_case_cross_namespace.php
@@ -1,0 +1,37 @@
+<?php
+
+// Namespace B is converted first: Holder's constant initializer forces the
+// lazy evaluation of A\E's case expression while B is the active conversion
+// namespace. `Helper::V` inside the case expression must resolve to A\Helper
+// (20), never to the decoy B\Helper (999).
+namespace B {
+    class Helper
+    {
+        public const V = 999;
+    }
+
+    class Holder
+    {
+        public const REF = \A\E::X;
+    }
+}
+
+namespace A {
+    class Helper
+    {
+        public const V = 20;
+    }
+
+    enum E: int
+    {
+        case X = Helper::V + 1;
+    }
+}
+
+namespace {
+    function main()
+    {
+        var_dump(\B\Holder::REF);
+        var_dump(\A\E::X->value);
+    }
+}
--- a/phpunit/code/enum_case_mutual_reference.php
+++ b/phpunit/code/enum_case_mutual_reference.php
@@ -1,0 +1,12 @@
+<?php
+
+enum E: int
+{
+    case A = E::B;
+    case B = E::A;
+}
+
+function main()
+{
+    var_dump(E::A);
+}
--- a/phpunit/code/enum_case_property_fetch.php
+++ b/phpunit/code/enum_case_property_fetch.php
@@ -1,0 +1,19 @@
+<?php
+
+enum E: int
+{
+    case One = 1;
+    case Two = E::One->value + 1;
+}
+
+enum S: string
+{
+    case A = 'a';
+    case B = S::A->name . '!';
+}
+
+function main()
+{
+    var_dump(E::Two->value);
+    var_dump(S::B->value);
+}
--- a/phpunit/code/enum_case_property_self_reference.php
+++ b/phpunit/code/enum_case_property_self_reference.php
@@ -1,0 +1,11 @@
+<?php
+
+enum G: int
+{
+    case A = G::A->value + 1;
+}
+
+function main()
+{
+    var_dump(G::A);
+}
--- a/phpunit/code/enum_case_self_reference.php
+++ b/phpunit/code/enum_case_self_reference.php
@@ -1,0 +1,11 @@
+<?php
+
+enum E: int
+{
+    case A = E::A;
+}
+
+function main()
+{
+    var_dump(E::A);
+}
--- a/phpunit/code/enum_case_sensitive_names.php
+++ b/phpunit/code/enum_case_sensitive_names.php
@@ -1,0 +1,25 @@
+<?php
+
+// Enum class names are case-insensitive, but case names are case-sensitive:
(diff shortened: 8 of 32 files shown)


matyhtf (changes requested):
Two enum paths are still outside these checks.

1. Forbidden magic methods introduced by a trait are accepted because the new check only runs in `prepareClassMethod()`. Composed and aliased trait methods are installed through `installComposedTraitMethod()`:

```php
trait T { public function __construct() {} }
enum Suit { use T; case Hearts; }
```

Zend reports `Enum Suit cannot include magic method __construct`; this PR dry-compiles it. Centralize the enum magic-method check and call it for composed/aliased methods as well. Add at least a trait-injected `__construct` test (an alias renamed to a forbidden magic name should follow the same path).

2. Backed enum case values are not limited to scalar literal nodes. `case Two = 1 + 1` and constant references are valid PHP, but this code still reads `$v->expr?->value` during prepare:

```php
enum Number: int { case Two = 1 + 1; }
function main(): void { var_dump(Number::Two->value); }
```

TypePHP emits `Undefined property: BinaryOp\Plus::$value`, records `null`, and later resolves the backed case as the string case name instead of integer `2`. Preserve the expression AST in prepare and evaluate it in the convert phase using the existing constant-expression machinery; do not assume a scalar node or evaluate implementation expressions during prepare. Add runtime coverage for an arithmetic expression and a constant reference.

AlessioGiacobbe:
Both paths closed, rebased on current master:

1. **Trait-composed magic methods** — the ban is centralized in `assertEnumMayIncludeMethod()` and now also runs where composed/aliased trait methods are installed, which sees post-adaptation names: both a trait-declared `__construct` and `use T { foo as __construct; }` fail with Zend's message ("Enum `Suit` cannot include magic method `__construct`"). Tests for both shapes.

2. **Non-literal backed case values** — prepare now stores scalar literals directly and keeps any other expression as its AST (no evaluation during prepare, no Throwable catching, no null overloading); the constant machinery evaluates it lazily on first convert-phase access and memoizes. Plain constant references resolve too, so `const TWO = 2; case Two = TWO;` works. `Number::Two->value` folds to `int(2)` instead of the case-name string, and the registration emits `ZVAL_LONG(..., 2)` (it always had the correct value — the compile-time folder was the broken half). Runtime PHPT covers arithmetic, constant references, a constant product, string concat, `->name`, and `from()`, with Zend-validated expectations.

The rebase also surfaced a pleasant interaction: upstream's abstract-method checks now fire before the enum complaint for `abstract private function` in enums, matching Zend's diagnostic order (probed).

matyhtf (changes requested):

Two constant-expression paths still need correction.

1. `enumCaseExprs` is unset only after evaluation succeeds, but there is no in-progress/visiting guard. Accessing a self-referencing case such as `enum E: int { case A = E::A; }` recursively enters `getClassConstValue()` until the compiler exhausts the stack/memory. Zend reports `Cannot declare self-referencing constant E::A`. Mutual cycles need the same protection.
2. Lazy evaluation can run while the translator is converting a different namespace. `getClassConstValue()` currently prefers `$this->namespace`, and the class-constant callback reduces `Name\FullyQualified` to a string, so an enum expression declared in namespace `A` may be resolved relative to namespace `B` when first referenced there.

Please add an evaluation stack keyed by the declaring enum/case, preserve the AST until successful completion, and resolve names using the expression declaration context/resolved AST name rather than the current conversion namespace. Add self-cycle, mutual-cycle, and cross-file/cross-namespace tests.

AlessioGiacobbe:
Rebased onto current master. Both paths fixed:

1. **Cycle guard** — lazy evaluation now runs inside an in-progress set keyed by `Enum\Fqn::CaseName`; re-entry emits `Cannot declare self-referencing constant \`E::A\``, and the AST stays in place until an evaluation completes successfully (popped in a finally). Probed against Zend 8.4.13 to match the reported constant exactly: `E::A` for `case A = E::A;`, `E::B` for the mutual pair `case A = E::B; case B = E::A;` regardless of which case is accessed, and likewise for a 3-cycle.
2. **Declaration-context resolution** — the class-constant callback now prefers the resolved AST name (`resolvedName` attribute / the node's fully qualified form), passed with a leading backslash so the active conversion namespace can never be prepended on top; and the preprocessor captures the declaring file's use-imports into `ClassDef::$enumUse*` (mirroring the existing `traitUse*` pattern), with evaluation running inside `withDeclarationNameContext()` — factored out of `withTraitNameContext()`, which now delegates to it.

Tests: self-cycle and mutual-cycle fatals with the exact messages, plus cross-namespace (with a decoy same-name symbol in the referencing namespace) and cross-file positives asserting the generated case/constant values, with the referencing file converted first.

matyhtf (changes requested):
Two compatibility gaps remain.

1. A valid backed-enum constant expression using an enum property fetch still fails:

```php
enum E: int {
    case One = 1;
    case Two = E::One->value + 1;
}
```

Zend PHP accepts this and resolves `E::Two->value` to `2`. This branch fails in the registration path because the independent constant-expression evaluator in `gen_stub.php` rejects `Expr_PropertyFetch`. Please either teach that evaluator about enum-case `->value` and `->name`, or pass the value already resolved by preprocessing into stub registration.

2. The cycle-detection key currently lowercases both the enum name and case name:

```php
$key = strtolower($enumName . "::" . $name);
```

Enum class names are case-insensitive, but enum case names are case-sensitive; PHP permits both `case A` and `case a`. Use the equivalent of:

```php
$key = strtolower($enumName) . "::" . $name;
```

Please add coverage for enum property fetches in backed values and for distinct `A` / `a` cases, including a dependency between their values.

AlessioGiacobbe:
Rebased onto current master; both gaps addressed.

1. Went with your second option — registration consumes the value the front-end already resolved — plus the prerequisite it implies: the front-end constant-expression evaluator also rejected `Expr_PropertyFetch`, so it now resolves `EnumClass::Case->value` / `->name` (object must itself be a constant expression evaluating to an enum case; property names case-sensitive). `EnumCaseInfo` carries the enum FQCN and `getDeclaration()` substitutes the translator-resolved scalar for any non-literal value expression before `EvaluatedValue::createFromExpression`, so gen_stub keeps a single source of truth instead of a second evaluator that could diverge. Cycle-reporting fidelity is preserved: registration enters unguarded (Zend's case fetch does not mark CONST_RECURSIVE either; only nested constant fetches mark), which keeps the mutual-cycle report at `E::B` for both access orders — re-probed on 8.4.13.

2. The cycle key now lowercases only the enum name: `strtolower($enumName) . '::' . $name`. Proven live — with the old key restored, the new fixture fails with a false self-reference on `F::a`.

Tests (Zend-verified): `E::One->value + 1` → `ZVAL_LONG(&enum_case_Two_value, 2)` (plus a string-backed `->name` concatenation case); distinct `A`/`a` cases with `a` derived from `A->value` and a forward reference between them; self-reference through `->value` reporting `Cannot declare self-referencing constant G::A` as Zend does.

matyhtf:
Thank you for identifying these enum declaration and constant-evaluation issues and for the detailed follow-up work.\n\nWe are going to close this PR because the remaining constant-cycle handling and integration conflicts need to be resolved together with the compiler constant-evaluation design on our side. We will fix the related problems directly in master and will reuse or adapt the valuable test cases and relevant implementation ideas from this PR.

matyhtf:
Additional context on the closure decision: PHP enums are runtime-backed class objects rather than C++-style compile-time constants. Deciding which enum semantics belong to preprocessing, constant folding, stub registration, and the Zend runtime affects the core TypePHP execution model.\n\nWe need to reconsider this from the language-design and compiler-architecture level instead of continuing with local fixes in this PR. The PR will therefore remain closed while we design and implement the solution internally. We will retain the reported cases, tests, and useful implementation ideas as references.

PR #62, opened 2026-09-01
fix(preprocessor): reject abstract method bodies and abstract private class methods

AlessioGiacobbe:
Abstract-method declaration rules were unenforced: an abstract method WITH a body compiled (the body silently dropped — Zend: "Abstract function A::f() cannot contain body", also fatal in traits, probed), and `abstract private function` in a class compiled (Zend fatal; legal in traits since PHP 8.0 — the trait exemption is preserved and tested).

Zend's diagnostic precedence (private before body) probed and matched.

Part of the split of #39.

### DIFF
--- a/phpunit/code/abstract_rule_body.php
+++ b/phpunit/code/abstract_rule_body.php
@@ -1,0 +1,4 @@
+<?php
+abstract class Job { abstract public function run(): void {} }
+
+function main() {}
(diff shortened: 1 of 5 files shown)


matyhtf (approved):
The declaration checks match Zend precedence for class and trait abstract methods. Focused tests pass (3/3). No blocking issue found.

PR #63, opened 2026-09-01
fix(preprocessor): enforce property-hook placement rules for class properties

AlessioGiacobbe:
Property-hook placement rules were unenforced on the class path (the interface path already had them): hooks on static properties ("Cannot declare hooks for static property"), hooks on readonly properties incl. readonly classes ("Hooked properties cannot be readonly"), abstract hooked properties in non-abstract classes, abstract properties without hooks, abstract hooked properties with a default or with all hooks bodied, and bodiless hooks on non-abstract properties (previously the lowering silently fabricated a concrete accessor).

Each rule and its precedence (static → readonly → abstract) probed against Zend 8.4.13; traits keep abstract hooked properties.

Part of the split of #39.

### DIFF
--- a/phpunit/code/hook_rule_abstract_all_bodies.php
+++ b/phpunit/code/hook_rule_abstract_all_bodies.php
@@ -1,0 +1,4 @@
+<?php
+abstract class Box { abstract public int $x { get => 1; } }
+
+function main() {}
--- a/phpunit/code/hook_rule_abstract_final.php
+++ b/phpunit/code/hook_rule_abstract_final.php
@@ -1,0 +1,4 @@
+<?php
+abstract class Box { abstract public int $x { final get; } }
+
+function main() {}
--- a/phpunit/code/hook_rule_abstract_no_hooks.php
+++ b/phpunit/code/hook_rule_abstract_no_hooks.php
@@ -1,0 +1,4 @@
+<?php
+abstract class Box { abstract public int $x; }
+
+function main() {}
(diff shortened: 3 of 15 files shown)


matyhtf (changes requested):
The abstract-property branch is missing two Zend modifier conflicts; both currently dry-compile:

```php
abstract class A {
    abstract private int $x { get; }
}
```

Zend: `Property hook cannot be both abstract and private`. Unlike abstract private trait methods, an abstract private property hook is also forbidden inside a trait. Protected abstract hooks remain legal.

```php
abstract class A {
    abstract public int $x { final get; }
}
```

Zend: `Property hook cannot be both abstract and final`. A bodiless hook must not carry `final`.

Please add these checks with the same Zend precedence and negative tests for both. The existing focused tests pass (8/8), but do not cover abstract/private or abstract/final combinations.

AlessioGiacobbe:
Added, with the precedence pinned by probes first: static → readonly → per-hook **final+private** ("Property hook cannot be both final and private" — it wins your combined `abstract private int $x { final get; }` case and also fires in traits and on body-carrying hooks) → then, for bodiless hooks, abstract+private and abstract+final, both of which fire in traits too (unlike abstract private trait *methods*) and precede the default-value and at-least-one-abstract-hook rules.

One probe finding worth noting: a `final` hook **with a body** on an abstract property is legal in Zend (`abstract public int $x { final get => 1; set; }` compiles), so the final check is strictly per-bodiless-hook.

Five new fixtures (abstract-private in class and trait, abstract-final, the final-private precedence case, and the protected-abstract positive), all Zend-validated; 27/27 across the hook suites.

PR #64, opened 2026-09-01
fix(preprocessor): reject variadic promoted properties and callable property/constant types

AlessioGiacobbe:
Two declaration gaps: variadic promoted constructor properties (`__construct(public int ...$x)`) compiled — the promotion was registered before the variadic check (Zend: "Cannot declare variadic promoted property") — and `callable` in property types (declared, promoted, union members) and class/interface constant types was accepted, where Zend rejects it ("Property A::$x cannot have type callable"; constants previously failed only incidentally with a confusing default-value error).

Messages include the full type string for composite types. Each rule probed against Zend 8.4.13.

Part of the split of #39.

### DIFF
--- a/phpunit/code/closure_rule_callable_intersection_param.php
+++ b/phpunit/code/closure_rule_callable_intersection_param.php
@@ -1,0 +1,5 @@
+<?php
+function main() {
+    $f = function (Traversable&callable $x): void {};
+    $f(null);
+}
--- a/phpunit/code/closure_rule_callable_intersection_return.php
+++ b/phpunit/code/closure_rule_callable_intersection_return.php
@@ -1,0 +1,5 @@
+<?php
+function main() {
+    $f = function (): Traversable&callable {};
+    $f();
+}
--- a/phpunit/code/const_rule_callable.php
+++ b/phpunit/code/const_rule_callable.php
@@ -1,0 +1,4 @@
+<?php
+class Bag { const callable FN = 1; }
+
+function main() {}
--- a/phpunit/code/const_rule_callable_dnf.php
+++ b/phpunit/code/const_rule_callable_dnf.php
@@ -1,0 +1,4 @@
+<?php
+class Bag { const (Traversable&callable)|int FN = 1; }
+
+function main() {}
--- a/phpunit/code/interface_rule_callable_dnf.php
+++ b/phpunit/code/interface_rule_callable_dnf.php
@@ -1,0 +1,4 @@
+<?php
+interface Baggy { public (Traversable&callable)|stdClass $fn { get; } const (Traversable&callable)|int FN = 1; }
+
+function main() {}
--- a/phpunit/code/param_rule_callable_dnf.php
+++ b/phpunit/code/param_rule_callable_dnf.php
@@ -1,0 +1,4 @@
+<?php
+function consume((Traversable&callable)|stdClass $value): void {}
+
+function main() {}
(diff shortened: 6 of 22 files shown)


matyhtf (changes requested):
`typeDeclContainsCallable()` deliberately returns `false` for every `IntersectionType`, but there is no merged fallback check on current `master`. A callable nested in DNF therefore reaches `gen_stub.php` and crashes on an assertion instead of producing a compiler diagnostic:

```php
class Bag {
    public (Traversable&callable)|stdClass $value;
}
```

Current result: `AssertionError: assert(!$type->isBuiltin)` in `StubType::getDnfTypeDeclarations()`. Zend rejects this at the declaration (`Type callable cannot be part of an intersection type`). The same gap applies to typed class constants and interface properties/constants.

Please make this PR safe independently: recurse into intersection members and reject `callable` with a stable fatal diagnostic, or rebase after the generic compound-type validator is merged and add explicit DNF callable tests proving the assertion is unreachable. The focused tests currently pass (5/5), but only cover bare and top-level union forms.

AlessioGiacobbe:
Fixed. Probing showed Zend rejects `callable` while compiling the intersection type itself, in every declaration context, with "Type callable cannot be part of an intersection type" — and that this wins over the property-specific ban (`callable|(Traversable&callable)` reports the intersection error). So instead of teaching `typeDeclContainsCallable()` to return the wrong (property) diagnostic for intersections, a dedicated walk over nullable/union/intersection members now runs *before* the existing callable checks at all four guard sites (class constants, declared + promoted properties, interface constants, interface properties).

Your exact case flips from the gen_stub `assert(!$type->isBuiltin)` crash to Zend's diagnostic; bare intersections, second-DNF-member, promoted, constant and interface variants are all covered, and a callable-free DNF (`(Countable&Traversable)|stdClass`) still compiles — proving the assertion is unreachable without over-rejection. 7 new fixtures, 31/31 focused tests green.

matyhtf (changes requested):

`assertTypeDeclIntersectionsHaveNoCallable()` is currently invoked only from property and class/interface constant paths. The same invalid type in a function parameter or return declaration still bypasses this check and can reach the later generator path:

```php
function consume(Traversable&callable $value): void {}
function produce(): Traversable&callable {}
```

Zend rejects both while compiling the type with `Type callable cannot be part of an intersection type`.

Please move this rule into the common type-declaration validation path used by parameters, returns, properties, promoted properties, class constants, interface members, and closures, rather than maintaining repeated per-context calls. Add negative parameter and return tests, including callable inside a DNF member.

AlessioGiacobbe:
Rebased onto current master. The rule now lives on the common path:

- `assertTypeDeclIntersectionsHaveNoCallable()` moved into `parseTypeDecl()` (behind `resolveTypeDecl()`), the funnel that parameters, returns, properties, promoted properties, class constants, interface constants and interface hooked properties all reach; the four per-context calls are removed. The diagnostic text matches Zend: `Type callable cannot be part of an intersection type`, anchored at the offending member node.
- Closures and arrow functions previously never resolved their full type nodes, so `doGenClosure()` now routes their parameter and return declarations through `resolveTypeDecl()` as well.
- The separate bare/nullable/union `callable` property/constant bans remain distinct rules (as in Zend), but `resolveTypeDecl()` is hoisted above them so `callable|(Traversable&callable)` reports the intersection conflict first, matching Zend's precedence.

Tests (each verified against Zend 8.4.13 first): negative parameter `Traversable&callable`, negative return, DNF member `(Traversable&callable)|stdClass` in a parameter, closure parameter and closure return; positive bare-`callable` parameter stays accepted.

PR #65, opened 2026-09-01
fix(preprocessor): validate compound type declarations and class-scope type keywords

AlessioGiacobbe:
Compound type declarations were not validated for well-formedness — all of these compiled, each a Zend compile fatal (probed individually): duplicate union members after alias/namespace resolution (`int|string|int`, `Foo|\Foo`, iterable expanded so `iterable|array` names the overlapping component); bool/true/false overlaps (`bool|false`, `true|false` → "bool must be used instead"); `mixed`/`void`/`never` inside unions; `?mixed`, `?null`, `?void`, `?never`; non-class standard types and duplicates inside intersections; `self`/`static` return types outside class scope (closures exempt, as Zend compiles them); duplicate `implements` entries for classes and enums.

One validation helper runs from a resolveTypeDecl override, so parameters, returns, properties, constants, and closures share the same pass without double-firing. `use T, T` is deliberately still accepted — Zend silently dedupes trait use (probed).

Part of the split of #39.

### DIFF
--- a/phpunit/code/type_rule_arrow_dup_union.php
+++ b/phpunit/code/type_rule_arrow_dup_union.php
@@ -1,0 +1,4 @@
+<?php
+function main(): void {
+    $fn = fn (int|string|int $value): int|string => $value;
+}
--- a/phpunit/code/type_rule_bool_false.php
+++ b/phpunit/code/type_rule_bool_false.php
@@ -1,0 +1,4 @@
+<?php
+function f(bool|false $x): void {}
+
+function main() {}
--- a/phpunit/code/type_rule_closure_callable_dnf.php
+++ b/phpunit/code/type_rule_closure_callable_dnf.php
@@ -1,0 +1,4 @@
+<?php
+function main(): void {
+    $closure = function ((Traversable&callable)|stdClass $value): void {};
+}
--- a/phpunit/code/type_rule_closure_dnf_permuted.php
+++ b/phpunit/code/type_rule_closure_dnf_permuted.php
@@ -1,0 +1,8 @@
+<?php
+interface A {}
+
+interface B {}
+
+function main(): void {
+    $closure = function ((A&B)|(B&A) $value): void {};
+}
--- a/phpunit/code/type_rule_closure_dup_union.php
+++ b/phpunit/code/type_rule_closure_dup_union.php
@@ -1,0 +1,4 @@
+<?php
+function main(): void {
+    $closure = function (int|string|int $value): void {};
+}
--- a/phpunit/code/type_rule_closure_intersect_scalar.php
+++ b/phpunit/code/type_rule_closure_intersect_scalar.php
@@ -1,0 +1,4 @@
+<?php
+function main(): void {
+    $closure = function (Countable&int $value): void {};
+}
--- a/phpunit/code/type_rule_closure_self_intersect_global.php
+++ b/phpunit/code/type_rule_closure_self_intersect_global.php
@@ -1,0 +1,4 @@
+<?php
+function main(): void {
+    $closure = function (self&Countable $value): void {};
+}
--- a/phpunit/code/type_rule_closure_self_return_valid.php
+++ b/phpunit/code/type_rule_closure_self_return_valid.php
@@ -1,0 +1,6 @@
+<?php
+function main(): void {
+    $closure = function (self $value): self {
+        throw new Exception('never bound');
+    };
+}
--- a/phpunit/code/type_rule_closure_static_return_valid.php
+++ b/phpunit/code/type_rule_closure_static_return_valid.php
@@ -1,0 +1,6 @@
+<?php
+function main(): void {
+    $closure = function (): static {
+        throw new Exception('never bound');
+    };
+}
--- a/phpunit/code/type_rule_dnf_permuted.php
+++ b/phpunit/code/type_rule_dnf_permuted.php
@@ -1,0 +1,4 @@
+<?php
+interface A {} interface B {} function f((A&B)|(B&A) $x): void {}
+
+function main() {}
--- a/phpunit/code/type_rule_dnf_subset.php
+++ b/phpunit/code/type_rule_dnf_subset.php
@@ -1,0 +1,4 @@
+<?php
+interface A {} interface B {} function f((A&B)|A $x): void {}
+
+function main() {}
--- a/phpunit/code/type_rule_dnf_subset_reversed.php
+++ b/phpunit/code/type_rule_dnf_subset_reversed.php
@@ -1,0 +1,4 @@
+<?php
+interface A {} interface B {} function f(A|(A&B) $x): void {}
+
+function main() {}
--- a/phpunit/code/type_rule_dnf_superset_group.php
+++ b/phpunit/code/type_rule_dnf_superset_group.php
@@ -1,0 +1,4 @@
+<?php
+interface A {} interface B {} interface C2 {} function f((A&B)|(A&B&C2) $x): void {}
+
+function main() {}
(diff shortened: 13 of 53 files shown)


matyhtf (changes requested):
The focused tests pass (14/14), but three Zend type rules still compile successfully:

1. `object` absorbs every class type:
```php
class Foo {}
function f(object|Foo $value): void {}
```
Zend: `Type Foo|object contains both object and a class type, which is redundant`. Track `object` versus resolved class members in both member orders.

2. Whole DNF groups cannot be skipped:
```php
interface A {}
interface B {}
function f((A&B)|(B&A) $value): void {}
```
Zend rejects the second intersection as redundant. It also rejects `(A&B)|A` because the intersection is more restrictive than `A`. Canonicalize intersection member sets (order-insensitive) and check exact/restrictive DNF redundancy.

3. Class-scope keywords are only checked for a bare return node. They also occur in parameters and nested union/DNF returns:
```php
function f(self $value): void {}
function g(): self|stdClass {}
```
Both must fail with `Cannot use "self" when no class scope is active`; this PR compiles them. Validate `self`/`static` recursively for every declaration context. `static` must additionally remain return-only.

Please add negative tests for these forms. The current comment explicitly saying whole-DNF redundancy is not checked documents a known PHP incompatibility, so this is not ready to merge as the compound-type validator.

AlessioGiacobbe:
All three rule sets implemented, rebased on current master (~40 fresh Zend 8.4.13 probes drove the exact messages and orderings):

1. **`object` + class-type redundancy** — fires in both member orders, with Zend's type rendering (class types in source order, builtins in Zend's canonical order); `iterable|object` and `null|object` stay legal as in Zend.
2. **Whole-DNF redundancy** — intersection member sets canonicalized order-insensitively: equal sets reject (``Type `B&A` is redundant with type `A&B` ``), and a strict superset rejects as more restrictive in both orders (`(A&B)|A`, `A|(A&B)`, `(A&B)|(A&B&C)`); `(A&B)|(A&C)` stays legal. The stale "not checked" comment is gone.
3. **Class-scope keywords everywhere** — `self`/`parent` validated recursively through nullable/union/intersection/DNF in parameters, returns, properties, and class/interface constants; closures stay exempt (Zend allows `self` there — verified). Two probe findings worth noting: `static` in a parameter or property type is a *parse* error in both Zend and php-parser (it can never reach the compiler), and Zend *accepts* `static` in class-constant types — so the only reachable `static` check is the no-class-scope return case, which was already exact. Documented in the commit body.

17 new fixtures (every negative one executed under real Zend and producing exactly the asserted error), 31/31 focused tests green, full suite and sweep identical to base.

matyhtf (changes requested):

The handling of `self`/`parent`/`static` inside intersections is not correct for DNF types. `validateIntersectionTypeDecl()` skips these names on the assumption that `buildTypeCheckFromNode()` rejects them later, but that rejection only runs when the top-level node is an `IntersectionType`; a DNF intersection nested inside a `UnionType` goes through `buildTypeCheckClause()` and is accepted.

For example, inside a class this branch accepts:

```php
class C {
    public function f((self&Countable)|stdClass $value): void {}
}
```

Zend rejects it with `Type self cannot be part of an intersection type`. The same needs coverage for `parent` in a class that actually has a parent. Note the PHP distinction: a bare `self&Countable` intersection is accepted, while `self` inside a DNF intersection is rejected, so this cannot be implemented as an unconditional ban in every intersection.

Please validate this at the compound-type layer with the necessary top-level/DNF context and add both positive bare-intersection and negative DNF tests.

AlessioGiacobbe:
Rebased onto current master and fixed at the compound layer — with one finding from probing PHP 8.4.13 directly: **Zend rejects class-scope keywords in every intersection, bare ones included**, so the bare/DNF distinction described in the review does not hold on 8.4:

```php
class C { public function f(self&Countable $v): void {} }
// Fatal error: Type self cannot be part of an intersection type
```

The shape Zend does accept is the keyword as a plain union member beside a DNF group — `(Ia&Ib)|self`, `(Ia&Ib)|parent`, `(Ia&Ib)|static` all compile. `parent`/`static` behave the same way in their intersections, and the scope-precedence messages win where applicable (`Cannot use "self" when no class scope is active`, `Cannot use "parent" when current class scope has no parent`).

Implementation follows the probed behavior: `validateIntersectionTypeDecl()` now rejects `self`/`parent`/`static` unconditionally with Zend's message (no context flag needed), the backstop in `buildTypeCheckFromNode()` reports identical text, and scope-precedence is preserved because `validateClassScopeTypeKeywords()` runs before the compound validator in every declaration context.

Tests: negatives for bare and DNF intersections across parameters, returns, properties, constants and promoted parameters (self, parent with a real parent, static in returns); positive coverage that `(Ia&Ib)|self|parent|static` union members still compile. Every fixture was first executed under real PHP 8.4.13. If 8.5 relaxes bare intersections here I'm happy to gate it, but 8.4 is unambiguous.

matyhtf (changes requested):
The compound-type validation is still limited to the preprocessing path. Closure and arrow-function signatures inside function bodies are resolved during conversion, so they bypass the `Preprocessor::resolveTypeDecl()` override.

For example, Zend rejects this duplicate union, but the current branch compiles it:

```php
function main(): void {
    $closure = function (int|string|int $value): void {};
}
```

Please move the compound-type well-formedness checks into a declaration path shared by preprocessing and conversion, or invoke the validator explicitly from closure/arrow generation. Keep the class-scope-keyword rule separate: a global closure may validly use `self` because it can later be bound to class scope.

Please add closure and arrow-function tests covering duplicate union members, scalar/callable members in intersections, and redundant DNF members.

AlessioGiacobbe:
Rebased onto current master and moved the compound well-formedness checks onto the shared path: `validateCompoundTypeDecl()` and its union/intersection helpers now live in `NameResolutionTrait` and run from `parseTypeDecl()` — the same funnel the merged callable-intersection rule uses — so preprocessing and conversion (closure/arrow signatures via `doGenClosure`) validate identically. The `Preprocessor::resolveTypeDecl()` override is gone.

`validateClassScopeTypeKeywords` stays per-context and closures never run it, matching the bindable-later semantics you described — probed on 8.4.13: a global closure with `: self` return, `self` param, or `: static` return compiles. One nuance the probes surfaced: `function (self&Countable $v)` in a global closure is still rejected by Zend with `Type self cannot be part of an intersection type`, so the keyword-in-intersection rule correctly stays on the shared path, separate from the scope rule.

Tests (all Zend-verified): closure and arrow-fn duplicate unions (`Duplicate type int is redundant`), scalar in intersection (`Type int cannot be part of an intersection type`), callable in a DNF member, permuted redundant DNF (`Type B&A is redundant with type A&B`), self-in-intersection global closure; positives for global closures declaring `self` and `static`.

Unrelated pre-existing gap noticed while probing (also on master, left out of scope here): a global closure with `: parent` return compiles in Zend but TypePHP rejects it with `Cannot use "parent" type declaration outside a class`.

matyhtf:
Thank you for identifying these compound-type validation issues and for the extensive PHP behavior probes.\n\nWe are going to close this PR because the remaining PHP 8.4 and PHP 8.5 compatibility differences need version-aware handling that we will implement directly on our side. We will reuse or adapt the valuable test cases and relevant implementation ideas from this PR.

matyhtf:
Additional context on the closure decision: compound-type validation affects the shared declaration pipeline, PHP-version-specific language rules, closure binding semantics, and the division of responsibility between php-parser, TypePHP, and Zend.\n\nThese are core language-design and compiler-architecture concerns, so we need to reconsider them as a coherent system rather than continue adding local validation rules in this PR. The PR will remain closed while we implement the solution internally. We will retain the reported cases, tests, and useful implementation ideas as references.

PR #66, opened 2026-09-01
fix: enforce readonly declaration and inheritance rules

AlessioGiacobbe:
Readonly declaration rules were unchecked for ZendVM-backed classes (checks existed only in the Native-class branch, which rejects readonly wholesale): `public readonly int $x = 5;` (Zend: "cannot have default value"), untyped readonly ("must have type"), and static readonly all compiled, for declared and promoted properties and via `readonly class` — each rule probed for exact Zend behavior; promoted readonly params keep their legal parameter defaults.

Readonly-class inheritance was also not sealed: a non-readonly class extending a readonly one (and the reverse) compiled; both directions now fail with Zend's messages.

One pre-existing fixture (inheritance_error_prop_readonly.php) used `readonly int $x = 2`, which Zend itself rejects before the inheritance error under test; the default was dropped so the fixture still exercises the mismatch.

Part of the split of #39.

### DIFF
--- a/phpunit/code/inheritance_error_prop_readonly.php
+++ b/phpunit/code/inheritance_error_prop_readonly.php
@@ -6,7 +6,7 @@ class A
 
 class B extends A
 {
-    public readonly int $x = 2;
+    public readonly int $x;
 }
 
 function main() {}
--- a/phpunit/code/readonly_class_allow_dynamic.php
+++ b/phpunit/code/readonly_class_allow_dynamic.php
@@ -1,0 +1,5 @@
+<?php
+#[AllowDynamicProperties]
+readonly class Cfg { public int $port; public function __construct() { $this->port = 80; } }
+
+function main() {}
--- a/phpunit/code/readonly_class_extends.php
+++ b/phpunit/code/readonly_class_extends.php
@@ -1,0 +1,5 @@
+<?php
+readonly class A {}
+class B extends A {}
+
+function main() {}
--- a/phpunit/code/readonly_class_extends_internal.php
+++ b/phpunit/code/readonly_class_extends_internal.php
@@ -1,0 +1,4 @@
+<?php
+readonly class Cfg extends ArrayObject {}
+
+function main() {}
--- a/phpunit/code/readonly_class_extends_rev.php
+++ b/phpunit/code/readonly_class_extends_rev.php
@@ -1,0 +1,5 @@
+<?php
+class A {}
+readonly class B extends A {}
+
+function main() {}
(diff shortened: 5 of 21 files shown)


matyhtf (changes requested):
The focused readonly tests pass (11/11), but three readonly class contracts still compile:

1. Trait property origin must retain readonlyness:
```php
trait T { public int $value; }
readonly class C { use T; }
```
Zend: `Readonly class C cannot use trait with a non-readonly property T::$value`. The current composition treats the consuming class flag as if it converted the trait property to readonly. It must reject the mismatch; add positive coverage for a trait property explicitly declared readonly.

2. Internal parents are skipped by the inheritance check:
```php
readonly class C extends ArrayObject {}
```
Zend: `Readonly class C cannot extend non-readonly class ArrayObject`. Query the internal parent through `ReflectionClass::isReadOnly()` (or equivalent metadata) so the rule applies to both TypePHP and internal parents.

3. `#[AllowDynamicProperties]` is forbidden on readonly classes:
```php
#[AllowDynamicProperties]
readonly class C { public function __construct(public int $value) {} }
```
Zend rejects it, while TypePHP compiles it. `tests/compiler/class/readonly-class.phpt` currently contains this invalid combination and should be corrected when adding the negative test.

Please cover and reject all three before merge.

AlessioGiacobbe:
All three covered, rebased on current master:

1. **Trait properties** — rejected during composition with Zend's message ("Readonly class `C` cannot use trait with a non-readonly property `T::$value`"). Placement detail: Zend names the *directly used* trait for nested traits (probed with an Inner→Outer chain — it reports `Outer::$v`), so the check runs in the composition pass where that name is in hand rather than off the origin attribute. Static trait properties are rejected too (they can never be readonly — probed); an explicitly-readonly trait property still composes (positive test).
2. **Internal parents** — the inheritance check no longer skips them: an internal-parent branch queries host `ReflectionClass::isReadOnly()`, two-directional in principle (today's only internal readonly classes, `BcMath\Number` and `Dom\NamespaceInfo`, are both final, so only the readonly-child direction is constructible — reflection keeps the reverse correct regardless). `readonly class C extends ArrayObject {}` now fails with Zend's message.
3. **`#[AllowDynamicProperties]`** — rejected on readonly classes with Zend's wording, and the invalid combination sitting in the pre-existing `readonly-class.phpt` is fixed (body re-validated under Zend, output unchanged); the negative test lives in the phpunit suite per the repo's convention for front-end rejections.

5 new fixtures, 21/21 with the related suites; sweep and full suite identical to base.

PR #69, opened 2026-09-02
fix(trait): preserve import context for property defaults

UEhQZXI:
I found this while trying to compile Laravel with TypePHP, in its `ParsesLogConfiguration` logging trait. The issue also reproduces without Laravel.

### Reproducer

Save as `repro.php`:

```php
<?php
namespace Repro\Support {
    enum Level: int { case Debug = 100; }
}

namespace Repro\Template {
    use Repro\Support\Level;

    trait HasLevels {
        public array $levels = ['debug' => Level::Debug];
    }
}

namespace Repro\App {
    class Logger {
        use \Repro\Template\HasLevels;
    }
}

namespace {
    function main(): void {
        echo (new \Repro\App\Logger())->levels['debug']->name, PHP_EOL;
    }
}
```

`php -r "require 'repro.php'; main();"` prints `Debug`.
`php bin/tpc.php repro.php --dry` fails with:

```text
Fatal error: Trait `Repro\Template\HasLevels` property `levels` conflicts with class `Repro\App\Logger`
```

### Fix

Composed properties were resolved with the consuming class's imports. Preserve their original trait context, including through nested traits, and use it when installing the properties. The consuming class scope and existing conflict checks are unchanged.

### Tests

- Add multi-file PHPUnit and nested-trait PHPT coverage, including conflicting consumer aliases.
- Verified PHP 8.4 output, the compiled Windows executable, and the Laravel logging scenario; 53/53 selected PHPT tests pass.
- Compared 1,540 PHPUnit tests and static-analysis diagnostics against the unmodified base: no new failures; existing baseline failures remain.

### DIFF
--- a/phpunit/code/trait-property-import-context/consumer.php
+++ b/phpunit/code/trait-property-import-context/consumer.php
@@ -1,0 +1,22 @@
+<?php
+
+namespace TraitPropertyImports;
+
+use TraitPropertyImports\Support\OtherLevel as ImportedLevel;
+
+trait NestedLevels
+{
+    use HasLevels;
+}
+
+class DirectConsumer
+{
+    use HasLevels;
+
+    public array $ownLevels = ['debug' => ImportedLevel::Debug];
+}
+
+class NestedConsumer
+{
+    use NestedLevels;
+}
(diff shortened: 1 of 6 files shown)


matyhtf (approved):
The origin marker survives nested trait flattening and `withTraitNameContext()` is applied only while installing the composed property, so the consumer scope is restored safely. Focused PHPUnit and the end-to-end PHPT both pass. No blocking issue found.

Issue #70, opened 2026-09-02
ArrayAccess 的 ??= miss 分支错误调用 offsetGet 而非 offsetSet，默认值未写入


hafung:
## Environment

- TypePHP: current `upstream/master` at `a6cd38a`
- Zend baselines: PHP 8.5.6 (Linux) and PHP 8.5.10 ZTS (Windows)
- AOT backend: Windows MSVC, freshly built and executed at both `-O0` and `-O3`

## Minimal reproducer

```php
<?php

declare(strict_types=1);

final class Bag implements ArrayAccess
{
    public array $data = [];

    public function offsetExists(mixed $offset): bool
    {
        echo "exists:$offset\n";
        return array_key_exists($offset, $this->data);
    }

    public function offsetGet(mixed $offset): mixed
    {
        echo "get:$offset\n";
        return $this->data[$offset] ?? null;
    }

    public function offsetSet(mixed $offset, mixed $value): void
    {
        echo "set:$offset=$value\n";
        $this->data[$offset] = $value;
    }

    public function offsetUnset(mixed $offset): void
    {
        unset($this->data[$offset]);
    }
}

function main(): void
{
    $bag = new Bag();
    $result = ($bag['service'] ??= 42);

    var_dump($result, $bag->data);
}
```

For Zend, invoke `main()` from a small runner. TypePHP binary mode invokes it directly.

## Expected (Zend 8.5.6 and 8.5.10)

```text
exists:service
set:service=42
int(42)
array(1) {
  ["service"]=>
  int(42)
}
```

## Actual TypePHP output (`-O0` and `-O3`)

```text
exists:service
get:service
int(42)
array(0) {
}
```

The assignment expression returns the RHS, but the default is not stored.

## Semantic invariant and failure mechanism

For an `ArrayAccess` target, `??=` must use Zend's dimension-operation sequence:

- missing offset: `offsetExists()` then `offsetSet()`;
- present non-null offset: `offsetExists()` then `offsetGet()`, without evaluating the RHS;
- `offsetExists() === true` and `offsetGet() === null`: evaluate the RHS and call `offsetSet()`.

The generated C++ for the reproducer is:

```cpp
result = (php::exists(bag, {{php::ArrayDimFetch, php::Var(get_str(2))}})
    ? bag.item(get_str(2), true)
    : (bag.item(get_str(2), true) = 42LL));
```

[...]

PR #71, opened 2026-09-02
fix(parser): preserve indirect ArrayAccess ??= semantics

hafung:
## Summary

Follow-up to #70 and the direct-variable fix in `4e13e7b`. This preserves PHP `??=` semantics when an array-dimension target may dispatch through `ArrayAccess` indirectly or change representation between phases.

- separate the presence/read phase from the write phase
- invoke `offsetExists()` once and `offsetGet()` only for a present offset
- dispatch a miss through the container that is current after RHS evaluation
- preserve the fixed-array lvalue fast path when no intervening expression can replace the container
- recursively preserve the same rules for nested `ArrayAccess` dimensions

## Failure mechanism and invariant

An `ArrayAccess` `offsetGet()` result is a value, not a writable bucket. In addition, a key expression, `offsetExists()`, or the RHS can mutate an aliased key/container. The read phase must use one stabilized receiver/key pair, while the write phase must resolve the current target after the RHS. Missing outer dimensions must not be created before the RHS.

The previous lowering handled a plain variable but did not enforce that invariant for nested dimensions, property/magic-property targets, references, or object/array representation changes.

## Coverage matrix

- direct, mixed, and referenced containers
- nested array-to-object and `ArrayAccess`-to-object targets
- ordinary and magic-property targets
- hit, null, and missing offsets
- key/container mutation from `offsetExists()` and from the RHS
- object-to-array and array-to-object rebinding
- result-used and result-unused paths (with the existing suite)
- fixed arrays retain their existing fast path unless an intervening operand can mutate the representation

Native objects and the features documented as incompatible remain outside this path.

## Verification

- reproducer and adversarial outputs compared exactly with Zend PHP 8.5
- `ArrayAccessCoalesceAssignCodegenTest`: 5 tests, 32 assertions
- target PHPT: 1/1 passed
- complete coalesce PHPT directory: 21/21 passed
- fresh Windows MSVC AOT artifacts built and executed at O0 and O3; both matched Zend output

### DIFF
--- a/phpunit/code/arrayaccess-coalesce-assign-codegen.php
+++ b/phpunit/code/arrayaccess-coalesce-assign-codegen.php
@@ -1,0 +1,53 @@
+<?php
+
+class CodegenArrayAccessBag implements ArrayAccess
+{
+    public function offsetExists(mixed $offset): bool { return false; }
+    public function offsetGet(mixed $offset): mixed { return null; }
+    public function offsetSet(mixed $offset, mixed $value): void {}
+    public function offsetUnset(mixed $offset): void {}
+}
+
+class CodegenArrayAccessHolder
+{
+    public function __construct(public mixed $value) {}
+
+    public function __get(string $name): mixed
+    {
+        return $this->value;
+    }
+}
+
+function coalesceArrayAccess(CodegenArrayAccessBag $bag, string $key): mixed
+{
+    return $bag[$key] ??= 42;
+}
+
+function coalesceMixedArrayAccess(mixed &$container, string $key): mixed
+{
+    return $container[$key] ??= 43;
+}
+
+function coalesceMagicArrayAccess(CodegenArrayAccessHolder $holder, string $key): mixed
+{
+    return $holder->virtual[$key] ??= 44;
+}
+
+function coalesceFixedArray(string $key): mixed
+{
+    $container = [];
+    return $container[$key] ??= 45;
+}
+
+function replaceCodegenArray(mixed &$container, string &$key): int
+{
+    $container = new CodegenArrayAccessBag();
+    $key = 'replacement';
+    return 46;
+}
+
+function coalesceMutableFixedArray(string $key): mixed
+{
+    $container = [];
+    return $container[$key] ??= replaceCodegenArray($container, $key);
+}
--- a/phpunit/src/ArrayAccessCoalesceAssignCodegenTest.php
+++ b/phpunit/src/ArrayAccessCoalesceAssignCodegenTest.php
@@ -1,0 +1,80 @@
+<?php
+
+use TypePhp\CompilerTest;
+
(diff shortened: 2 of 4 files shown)


matyhtf (changes requested):

The write phase currently dispatches only as `isArray() ? item(..., true) : offsetSet(...)`. The second branch is not equivalent to Zend dimension assignment for a dynamic `mixed`/`php::Var` container. For example:

```php
function f(mixed $value): void {
    $value['k'] ??= 7;
}
f(1);
```

Zend throws `Cannot use a scalar value as an array`. `Variant::offsetSet()` currently does nothing for an integer, so this lowering can return `7` while leaving the target unchanged. Other runtime representations also differ: null must become an array, false follows PHP's deprecated conversion behavior, invalid scalars must throw, and string offsets have their own key/range errors.

Please route the write through a PHPX helper that implements Zend write-dimension semantics for the complete runtime value domain, while retaining the explicit ArrayAccess path only after proving the runtime value is an object. Add PHPT coverage for dynamic null, false, int/float/bool, string keys/offsets, arrays, references, and ArrayAccess objects.

hafung:
Addressed the review in bd6b8be and rebased onto current master (e90d624).

Changes:
- dynamic non-object writes now go through a Zend-compatible PHPX helper instead of Variant::offsetSet()
- the explicit offsetSet() path is selected only after proving the runtime receiver is an object
- assignment expressions return the actual write result (important for string-offset writes)
- PHPT coverage now includes dynamic null, false, true/int/float, strings (valid/invalid keys and extension), arrays, references/aliases, and ArrayAccess

The required PHPX helper is proposed in swoole/phpx#102.

Verification completed locally:
- Zend reproducer/output comparison: pass
- targeted PHPUnit codegen suite: 5 tests, 32 assertions
- compiler coalesce PHPT suite: 21/21 pass
- fresh AOT O0 artifact: pass
- fresh AOT O3 artifact: pass
- generated code inspected for receiver/key/value ordering and object-vs-nonobject dispatch

hafung:
Follow-up for the null-key/append distinction is now pushed in `5518206` (PHPX prerequisite updated to `dd802ee`).

- Renamed the runtime API to `assignKeyedDimension()` so it explicitly represents a syntactically present dimension; it is not an append API.
- Added a dynamic-array regression proving that a `null` key emits Zend's deprecation and writes the empty-string key.
- Added an ArrayAccess regression proving that `offsetExists()` and `offsetSet()` still receive the original `null` key. PHP's `$object[]` ambiguity remains confined to ArrayAccess and is not used to infer array append semantics.

Reverified the targeted PHPUnit suite (5 tests, 32 assertions), the PHPT, and fresh AOT O0/O3 artifacts against PHP 8.5.10; all pass with matching output.

matyhtf:
https://github.com/swoole/typephp/commit/4b722427496fc5e4addd8900512736a727883379

PR #77, opened 2026-09-03
fix(optimizer): preserve explicit round mode for Decimal

### LINKED ISSUE #74: optimizer: round() drops an explicit mode for Decimal inputs
## Reproducer

```php
<?php

function main(): void
{
    $decimal = std::decimal('2.5');
    var_dump(round($decimal, 0, PHP_ROUND_HALF_DOWN));
}
```

On current `master`, `genRound()` selects the Decimal fast path before it checks for a third argument. The generated C++ is equivalent to:

```cpp
php::Decimal::round(decimal, 0L)
```

The explicitly supplied rounding mode is silently discarded.

## Invariant and failure mechanism

An optimizer must not select an implementation that cannot represent all explicitly supplied arguments. `php::Decimal::round()` supports the one- and two-argument extension forms, but it has no rounding-mode parameter. A three-argument call must remain on the generic runtime path, which preserves normal argument validation.

PR #30 routes explicit `round()` modes away from `php::fn::round()`, but its arity check currently occurs after the Decimal early return, leaving this alternate dispatch path uncovered.

## Relevant matrix

- Decimal with one or two arguments: retain the existing fast path.
- Decimal with an explicit third argument: generic runtime path, regardless of whether the mode is a legacy integer, `RoundingMode`, or invalid value.
- int/float with an explicit mode: already covered by #30 and must remain generic.
- named and unpacked arguments: already rejected before positional lowering by #30.
- result-used/unused contexts do not alter dispatch; both use the same call expression lowering.

The code-generation reproducer on current `master` reports three `php::Decimal::round()` calls for the one-, two-, and three-argument forms; the third must instead be a `php::call()`.

hafung:
Fixes #74

## Summary

This is a follow-up to #30. Move the explicit-mode guard in `genRound()` ahead of both optimized implementations so a three-argument Decimal call cannot silently discard its mode.

The one- and two-argument numeric and Decimal fast paths are unchanged. Named and unpacked arguments retain #30's existing early fallback.

## Invariant and failure mechanism

An optimizer may select a fast path only when it can represent every explicitly supplied argument. `php::Decimal::round()` supports TypePHP's one- and two-argument Decimal extension, but has no mode parameter.

On unpatched current `master`, the code-generation reproducer emits three `php::Decimal::round()` calls for these forms:

```php
round($decimal);
round($decimal, 2);
round($decimal, 0, PHP_ROUND_HALF_DOWN);
```

The last call is lowered as `php::Decimal::round($decimal, 0)` and loses the mode. The shared arity guard now rejects it before container/type dispatch, leaving the generic runtime call to validate the arguments.

## Relevance matrix

- Decimal, one/two arguments: native Decimal path retained.
- Decimal, explicit mode: generic path for legacy integers, `RoundingMode`, and invalid values.
- int/float, explicit mode: remains generic as established by #30.
- int/float, one/two arguments: PHPX fast path retained.
- named/unpacked calls: existing generic fallback retained.
- used/unused results share the same call-expression dispatch, so no separate lowering path is involved.
- PHPX 2.6.11 now permits the runtime PHPT to cover an internal `RoundingMode` case that #30 had to omit.

## Verification

- Unpatched current upstream: new codegen assertion fails with 3 Decimal fast-path calls instead of 2.
- `phpunit/src/RoundModeTest.php`: 5 tests, 8 assertions.
- Windows PHP 8.5.10/MSVC target PHPT: 1/1 passed.
- TypePHP self-host build completed with the patched optimizer (`/Od`).
- Fresh target artifacts compiled and executed in distinct paths at O0 (`/Od`) and O3 (`/Ox`); complete outputs were byte-identical and matched the PHPT expectation.
- Generated C++ inspected: the one/two-argument Decimal calls use `php::Decimal::round`, while the explicit-mode call uses `php::call` with all three operands.

### DIFF
--- a/phpunit/code/round-with-decimal-mode.php
+++ b/phpunit/code/round-with-decimal-mode.php
@@ -1,0 +1,16 @@
+<?php
+/**
+ * This file is part of TypePHP(AOT).
+ *
+ * @link     https://www.swoole.com/aot/
+ * @contact  service@swoole.com
+ */
+
+function main(): void
+{
+    $decimal = std::decimal('2.5');
+
+    var_dump(round($decimal));
+    var_dump(round($decimal, 2));
+    var_dump(round($decimal, 0, PHP_ROUND_HALF_DOWN));
+}
--- a/phpunit/src/RoundModeTest.php
+++ b/phpunit/src/RoundModeTest.php
@@ -56,6 +56,16 @@ class RoundModeTest extends TestCase
         self::assertSame(2, substr_count($cpp, 'php::fn::round('));
     }
 
+    public function testDecimalWithAnExplicitModeReachesTheRuntime(): void
+    {
+        $cpp = $this->compileToCpp('round-with-decimal-mode.php');
+
+        self::assertSame(2, substr_count($cpp, 'php::Decimal::round('));
+        // Three calls wrap the results in var_dump(); the fourth is round()
+        // itself after the explicit mode disqualifies the Decimal fast path.
+        self::assertSame(4, substr_count($cpp, 'php::call('));
+    }
+
     private function compileToCpp(string $file): string
     {
         global $translator;
--- a/src/Optimizer/FuncCallOptimizer.php
+++ b/src/Optimizer/FuncCallOptimizer.php
@@ -1013,10 +1013,17 @@ trait FuncCallOptimizer
                 return false;
             }
         }
+        $args = count($e->args);
+        if ($args >= 3) {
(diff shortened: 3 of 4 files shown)


hafung:
Superseded by 643e49d56ef6ffaf9d46dbe095b5d8da850da871, which fixes #74 at the intended compatibility boundary by rejecting unsupported Decimal modes explicitly while preserving the one/two-argument fast path. Closing this alternate dynamic-fallback approach.

PR #79, opened 2026-09-03
optimize: convert for-loop post-inc/dec to prefix to avoid zval copy

yuan-dian:
## Summary

For-loop post-expressions (`$i++`, `$i--`) are converted to prefix form (`++$i`, `--$i`) when the return value is discarded. This eliminates the zval copy + destructor overhead from `operator++(int)`'s return value.

## Motivation

In C++, postfix `operator++(int)` must copy the zval to return the old value, then destroy the temporary. Prefix `operator++()` modifies in-place with no copy. In tight loops, this saves ~6ns per increment.

Benchmark (n=1M, rounds=3, -O3 -flto):

| Scenario | Before | After | Change |
|----------|--------|-------|--------|
| Loop Only (count++) | 19.06 ms | 4.53 ms | ↓76% |
| Loop + Compare (==) | 18.37 ms | 3.44 ms | ↓81% |

## Changes

- **`src/Parser/LoopControlTrait.php`**: Convert `PostInc` → `PreInc`, `PostDec` → `PreDec` in for-loop post-expressions only (third clause of `for (init; cond; post)`)
- **`phpunit/src/LoopControlTest.php`**: Add `testForLoopPostIncConvertedToPreInc` verifying generated C++ contains `++i`/`--i` for for-loops and retains `i++` in while/do-while bodies
- **`phpunit/code/loop/for-loop-post-expr-opt.php`**: Test data with 7 functions covering basic, multi-post, nested, mixed, and while/do-while cases

## Safety

- Only affects for-loop post-expressions — while/do-while body `i++` untouched
- `$i++` ≡ `++$i` when return value is discarded (C++ standard)
- `LoopVarOptimizer::detectLoopUnitStep()` already handles both `PreInc` and `PostInc`

### DIFF
--- a/phpunit/code/loop/for-loop-post-expr-opt.php
+++ b/phpunit/code/loop/for-loop-post-expr-opt.php
@@ -1,0 +1,47 @@
+<?php
+
+function test_post_inc_basic(): void {
+    $sum = 0;
+    for ($i = 0; $i < 10; $i++) {
+        $sum += $i;
+    }
+}
+
+function test_post_dec_basic(): void {
+    $sum = 0;
+    for ($i = 10; $i > 0; $i--) {
+        $sum += $i;
+    }
+}
+
+function test_multi_post(): void {
+    for ($i = 0, $j = 10; $i < 5; $i++, $j--) {
+    }
+}
+
+function test_mixed_post(): void {
+    $j = 0;
+    for ($i = 0; $i < 10; $i++, $j += 2) {
+    }
+}
+
+function test_nested_for(): void {
+    for ($i = 0; $i < 5; $i++) {
+        for ($j = 0; $j < 5; $j++) {
+        }
+    }
+}
+
+function test_while_not_affected(): void {
+    $i = 0;
+    while ($i < 10) {
+        $i++;
+    }
+}
+
+function test_do_while_not_affected(): void {
+    $i = 0;
+    do {
+        $i++;
+    } while ($i < 10);
+}
--- a/phpunit/code/loop/for-loop-post-expr-opt.php
+++ b/phpunit/code/loop/for-loop-post-expr-opt.php
@@ -45,3 +45,53 @@ function test_do_while_not_affected(): void {
         $i++;
     } while ($i < 10);
 }
+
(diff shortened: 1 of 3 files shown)


matyhtf:
Please restrict this optimization to ordinary variable expressions only:

```php
if ($expr instanceof Node\Expr\PostInc && $this->isVarExpr($expr->var)) {
    $expr = new Node\Expr\PreInc($expr->var, $expr->getAttributes());
} elseif ($expr instanceof Node\Expr\PostDec && $this->isVarExpr($expr->var)) {
    $expr = new Node\Expr\PreDec($expr->var, $expr->getAttributes());
}
```

There is no need to add a separate `is_string($expr->var->name)` check. Variable variables are already rejected with a fatal error by `parseVariable()` when the variable name is not a string.

The current rewrite also accepts properties, static properties, and array elements. TypePHP does not lower prefix and postfix operations identically for every compound lvalue. In particular, a dynamic static-property postfix operation uses an explicit get/set write-back path, while rewriting it to prefix may only increment the temporary value returned by `getStaticProperty()`.

Please also add tests confirming that property, static-property, and array-element postfix operations are not rewritten by this optimization.

yuan-dian:
Thanks for the guidance. Updated to restrict the optimization to simple variable expressions only using isVarExpr():
```php
if ($expr instanceof Node\Expr\PostInc && $this->isVarExpr($expr->var)) {
    $expr = new Node\Expr\PreInc($expr->var, $expr->getAttributes());
} elseif ($expr instanceof Node\Expr\PostDec && $this->isVarExpr($expr->var)) {
    $expr = new Node\Expr\PreDec($expr->var, $expr->getAttributes());
}
```
Added 6 test cases confirming that property, static-property, and array-element postfix operations are not rewritten:
- test_property_post_inc_not_rewritten / test_property_post_dec_not_rewritten
- test_static_property_post_inc_not_rewritten / test_static_property_post_dec_not_rewritten
- test_array_element_post_inc_not_rewritten / test_array_element_post_dec_not_rewritten

All tests pass (5 tests, 13 assertions).

PR #81, opened 2026-09-03
perf: optimize checked integer returns

hafung:
## Summary

- add a checked native fast path for `int + int`, `int - int`, and `int * int` at an exact `: int` return boundary
- preserve ordinary PHP overflow semantics: the uncommon overflow branch still materializes the widened `float` value and raises the existing return `TypeError`
- conservatively keep the current Variant lowering for nested/side-effecting operands, closures, composite returns, and native-types mode
- use the checked-in PHPX 2.6.12 overflow primitives; this needs no PHPX change or dependency update

## Why this is safe for TypePHP's hybrid model

Ordinary PHP integer arithmetic is not statically `int`: overflow widens the result to `float`. However, at an exact `: int` return boundary, that widened value cannot escape; it can only fail the existing strict return check.

The generated code therefore speculates on the common non-overflowing case with checked native arithmetic. On overflow it reconstructs the PHP float result and routes it through the normal TypePHP return-type diagnostic. The optimization is restricted to reusable typed-int variables and foldable integer constants so neither operand is evaluated twice.

| Dimension | Lowering |
| --- | --- |
| exact `: int`, reusable int operands, `+`/`-`/`*` | checked native fast path |
| runtime overflow | float Variant + existing `TypeError` path |
| nested/call/property/side-effecting operands | existing dynamic fallback |
| closure or union/nullable/composite return | existing dynamic fallback |
| `use native_types` / explicit native arithmetic | existing native lowering |
| other operators or operand types | unchanged |

This intentionally optimizes the static fact available at the return boundary without treating the program as wholly static.

## Performance

Fresh upstream and patched artifacts were built with the same PHP 8.5.6 ZTS / Release PHPX toolchain, GCC `-O3 -flto`. Each number is the median of 7 independent process runs; each run reports the best of 5 rounds over 10,000,000 calls.

| Benchmark | upstream/master | patched | Reduction | Speedup |
| --- | ---: | ---: | ---: | ---: |
| typed function call (`addInts`) | 17.131 ns/op | 1.145 ns/op | 93.3% | 14.96x |
| typed method call | 16.994 ns/op | 1.145 ns/op | 93.3% | 14.84x |
| function call with iteration count parsed from `argv` | 17.063 ns/op | 1.140 ns/op | 93.3% | 14.97x |

The large gain comes from removing the boxed return on the hot non-overflowing path, which lets the C++ optimizer inline and optimize through the function/method boundary. The runtime-input case guards against relying on a compile-time-known loop count.

### Zend PHP comparison

The same source was also run with the matching PHP 8.5.6 ZTS CLI, both with Opcache/JIT disabled and with CLI Opcache plus tracing JIT (`opcache.jit=tracing`, 128 MiB JIT buffer). These are medians of 7 independent processes, with the same warm-up and inner-round policy.

| Benchmark | Zend, no JIT | Zend tracing JIT | TypePHP O3+LTO | TypePHP vs no JIT | TypePHP vs JIT |
| --- | ---: | ---: | ---: | ---: | ---: |
| typed function call | 9.918 ns/op | 2.857 ns/op | 1.143 ns/op | 8.68x faster | 2.50x faster |
| typed method call | 11.194 ns/op | 3.047 ns/op | 1.147 ns/op | 9.76x faster | 2.66x faster |
| runtime-input function call | 9.906 ns/op | 2.857 ns/op | 1.141 ns/op | 8.68x faster | 2.50x faster |

This is a narrow arithmetic-call microbenchmark, not an application-wide claim. The maintained full bridge matrix still shows slower dynamic paths (notably magic calls/properties and string concatenation), which are separate optimization targets.

## Validation

- full PHPUnit: 1760 tests, 4557 assertions (pass; existing warnings/deprecations remain)
- targeted codegen PHPUnit: 4 tests, 16 assertions
- target overflow PHPT passes, including normal add/subtract/multiply and all three overflow `TypeError` cases
- self-host compiler build succeeds; `ldd` resolves the checkout PHPX and `/home/zz/.typephp` PHP; `./tpc --version` succeeds
- fresh O0 and O3+LTO AOT artifacts both match Zend output exactly; generated C++ inspected in both builds
- operator + optimization PHPT group: 83/85 pass; the two failures are unchanged on pristine upstream/master and are PHP 8.5 implicit float-to-int deprecation output (`float-bitwise-mod.phpt`, `float-narrow-nested-int-ops.phpt`)

## Relation to #79

#79 only rewrites unused for-loop post-inc/dec expressions to prefix form in `LoopControlTrait`. This PR changes checked arithmetic at exact integer return boundaries in `BinaryOpTrait`/return lowering. The files, semantic invariant, and hot paths do not overlap.

### DIFF
--- a/phpunit/code/typed-scalar-arithmetic-codegen.php
+++ b/phpunit/code/typed-scalar-arithmetic-codegen.php
@@ -1,5 +1,25 @@
 <?php
 
+function addTypedInts(int $a, int $b): int
+{
+    return $a + $b;
+}
+
+function subtractTypedInts(int $a, int $b): int
+{
+    return $a - $b;
+}
+
+function multiplyTypedInts(int $a, int $b): int
+{
+    return $a * $b;
+}
+
+function addNestedTypedInts(int $a, int $b): int
+{
+    return ($a + $b) + 1;
+}
+
 function divTypedInts(int $a, int $b): float
 {
     return $a / $b;
--- a/phpunit/src/TypedScalarArithmeticCodegenTest.php
+++ b/phpunit/src/TypedScalarArithmeticCodegenTest.php
@@ -11,6 +11,16 @@ use TypePhp\CompilerTest;
  */
 final class TypedScalarArithmeticCodegenTest extends \BaseTest
 {
+    public function testTypedIntReturnUsesCheckedNativeFastPath(): void
+    {
+        $code = $this->compileFixture();
+
+        self::assertSame(1, substr_count($code, 'php::detail::intAddOverflow('));
+        self::assertSame(1, substr_count($code, 'php::detail::intSubOverflow('));
+        self::assertSame(1, substr_count($code, 'php::detail::intMulOverflow('));
+        self::assertStringContainsString('((php::Var(a)) + (b))', $code);
+    }
+
     public function testTypedIntDivisionRoutesThroughVariant(): void
     {
         $code = $this->compileFixture();
--- a/src/CompilerBase.php
+++ b/src/CompilerBase.php
@@ -2535,6 +2535,10 @@ class CompilerBase implements PropertyAccessContext
             $this->context->afterStmtLines[] = $this->getIndent() . 'return ' . $finalReturn . ';';
             return $tmpVar . ' = ' . $returnExpr . ';';
         }
+        $checkedIntReturn = $this->tryParseCheckedIntArithmeticReturn($v->expr);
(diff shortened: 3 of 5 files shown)


matyhtf:
这种优化没必要，直接用 use native_types 就好了，这种简单的 四则运算，只是刷分，实际场景没用。 而且 php::detail:: 还是内部方法，不能用在 typephp 里面。

PR #82, opened 2026-09-03
fix(trait): #80 finalize composed trait methods before file conversion

### LINKED ISSUE #80: 跨文件 Trait 方法调用降级为 `__call()` 导致运行时错误
## 摘要

TypePHP 编译器在编译跨文件调用 trait 方法时，无法正确解析方法来源，将直接调用降级为 `__call()` 魔术方法调用。当目标类的 `__call()` 转发给底层不具有该方法的对象时，运行时报错。

## 错误信息

```
Error: Invalid callback ReflectionClass::getAttribute,
class ReflectionClass does not have a method "getAttribute"
```

### 文件 1: `src/TraitA.php`

```php
<?php
namespace Lib;

trait TraitA
{
    abstract public function getRefl(): \ReflectionClass;

    public function getAttribute(string $cls): ?object
    {
        $attrs = $this->getRefl()->getAttributes($cls);
        return $attrs[0] ?? null ? $attrs[0]->newInstance() : null;
    }
}
```

### 文件 2: `src/ClassB.php`

```php
<?php
namespace Lib;

use ReflectionClass;

class ClassB
{
    use TraitA;

    private ReflectionClass $ref;

    public function __construct(string $class)
    {
        $this->ref = new ReflectionClass($class);
    }

    public function getRefl(): ReflectionClass
    {
        return $this->ref;
    }

    public function __call(string $name, array $args): mixed
    {
        return $this->ref->$name(...$args);
    }
}
```

### 文件 3: `main.php`

```php
<?php
function main(): void
{
    require __DIR__ . '/src/TraitA.php';
    require __DIR__ . '/src/ClassB.php';

    $obj = new \Lib\ClassB(\SomeClass::class);
    $result = $obj->getAttribute(\SomeAttr::class);
    echo $result?->name ?? 'not found';
}
```

## 预期行为

编译器应生成直接调用 trait 方法的 C++ 代码：

```cpp
result = php_lib__classb__getattribute(obj, arg);
```

## 实际行为

编译器生成降级为 `__call()` 的调用：

```cpp
result = php_lib__classb____call(obj, "getAttribute", {arg});
```

`__call()` 转发给 `ReflectionClass`，而 `ReflectionClass` 没有 `getAttribute()` 方法，运行时报错。

### 实际生成（错误）

```cpp
// 编译器生成
attr = php_lib__classb____call(ref, "getAttribute", args);
//                      ^^^^ 降级为 __call
```

### 预期生成（正确）

```cpp
// 应该生成
attr = php_lib__classb__getattribute(ref, args);
//                      ^^^^ 直接调用 trait 方法
```

## 根本原因

编译器在 `MethodCallTrait::parseMethodCall()` 中无法从 trait 定义中识别方法，当类存在 `__call()` 时降级为动态调用。

## 关键代码位置

| 位置 | 说明 |
|------|------|
| `Parser/MethodCallTrait.php:517` | 方法调用解析入口 `parseMethodCall()` |
| `CompilerBase.php:2764` | 方法查找 `getNativeMethod()` |
| `Translator.php:3164` | Trait AST 组合 `composeTraitAst()` |
| `Parser/MethodCallTrait.php:767-784` | `__call` 回退逻辑 |

## 影响范围

所有满足以下条件的场景：

1. 类使用 trait 提供方法
2. 同时定义 `__call()` 魔术方法
3. 外部文件调用该 trait 方法

prateekbhujel:
## Summary

Fixes #80.

When compiling multi-file projects where a class uses a trait and declares `__call()`, calling the trait method from an external file that happens to be converted before the consuming class failed to locate the method in the class symbol table because trait methods were only composed during `parseClass()` of the consuming class itself during its file conversion.

As a result, `MethodCallTrait::findNativeMethod()` could not find the method on the consuming class, saw that `__call()` was defined, and threw `DynamicCall`, which was lowered into `php_<class>____call(...)`. If `__call()` delegates dynamically to an underlying object lacking that method, a runtime error is raised.

## Solution

1. **Trait Resolution in `CompilerBase::getNativeMethod()`**:
   When resolving a method on a class, if the method is not directly declared on the class body, check `findComposedTraitMethod()` across the class's `usedTraits` hierarchy. When found, cache the discovered function in `CompilerBase::$traitMethodFunctions`.
2. **Transparent Function Resolution**:
   In `CompilationStateTrait::hasFunction()` and `getFunction()`, check `$this->traitMethodFunctions` if the function is not yet in the symbol table. When the consuming class is subsequently converted, `addFunction()` unsets the temporary entry and replaces it with the permanent function definition in `SymbolRepository`.
3. **Trait Interface Satisfaction in `Translator::findClassMethodDef()`**:
   Check `findComposedTraitMethod()` when validating interface method implementations so that trait-provided methods correctly satisfy interface contracts across translation units.

## Tests

- Added `phpunit/src/TraitCrossFileCallTest.php` asserting that `php_lib__classb__getattribute(...)` is directly generated without falling back to `__call()` or dynamic dispatch regardless of file conversion order.
- Added PHPT test `tests/compiler/trait/trait-cross-file-method-magic-call.phpt`.
- All 40 Trait PHPUnit tests pass 100%.
- Full 166-file self-hosting build and conversion passes with 0 errors.

### DIFF
--- a/phpunit/code/trait-cross-file-call/caller.php
+++ b/phpunit/code/trait-cross-file-call/caller.php
@@ -1,0 +1,9 @@
+<?php
+
+use Lib\ClassB;
+
+function callTraitMethod(): void
+{
+    $b = new ClassB();
+    $b->getAttribute('SomeClass');
+}
--- a/phpunit/code/trait-cross-file-call/class.php
+++ b/phpunit/code/trait-cross-file-call/class.php
@@ -1,0 +1,13 @@
+<?php
+
+namespace Lib;
+
+class ClassB
+{
+    use TraitA;
+
+    public function __call(string $name, array $args): mixed
+    {
+        return null;
+    }
+}
--- a/phpunit/code/trait-cross-file-call/trait.php
+++ b/phpunit/code/trait-cross-file-call/trait.php
@@ -1,0 +1,11 @@
+<?php
+
+namespace Lib;
+
+trait TraitA
+{
+    public function getAttribute(string $cls): string
+    {
+        return $cls;
+    }
+}
--- a/phpunit/src/TraitCrossFileCallTest.php
+++ b/phpunit/src/TraitCrossFileCallTest.php
@@ -1,0 +1,49 @@
+<?php
+/**
+ * This file is part of TypePHP(AOT).
+ *
+ * @link     https://www.swoole.com/aot/
+ * @contact  service@swoole.com
+ */
+
+use TypePhp\CompilerTest;
+
+/**
+ * @internal
+ * @coversNothing
+ */
+final class TraitCrossFileCallTest extends BaseTest
+{
+    public function testCrossFileTraitMethodCallEmitsDirectNativeCallRegardlessOfConversionOrder(): void
+    {
+        global $translator;
+
+        $compiler = CompilerTest::create(TYPEPHP_ROOT_PATH);
+        $translator = $compiler;
+        $directory = TYPEPHP_ROOT_PATH . '/phpunit/code/trait-cross-file-call/';
+        $traitFile = $directory . 'trait.php';
+        $classFile = $directory . 'class.php';
+        $callerFile = $directory . 'caller.php';
+
+        $compiler->addFiles([$traitFile, $classFile, $callerFile]);
+        $compiler->prepareFile($traitFile);
+        $compiler->prepareFile($classFile);
+        $compiler->prepareFile($callerFile);
+
+        // Convert caller.php BEFORE class.php to verify order independence
+        $callerGenerated = $compiler->convertFile($callerFile);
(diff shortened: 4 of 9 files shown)


matyhtf:
This issue has now been fixed on master in commit d1a9397f. The compiler now uses an explicit three-stage flow: prepare all declarations, compose and inject Trait declarations after the complete symbol graph is available, and only then convert method implementations.

The approach in this PR does not match the TypePHP compiler architecture: Trait declarations cannot be expanded during per-file preparation because the referenced Trait may not have been prepared yet, while delaying declaration injection until conversion is too late for callers compiled earlier. The PR also changes ClassDef and the compilation pipeline more broadly than necessary.

Closing this PR because the underlying bug has been fixed with the declaration-composition stage.

PR #86, opened 2026-09-05
fix(parser): preserve numeric variable types (#85)

### LINKED ISSUE #85: without `use native_types`numeric variable stored as php::Var
Consider the following code:

```php
<?php
declare(strict_types=1);

function main()
{
    $x = 42;
    $x = [1, 2];
}
```

This compiles without any error and value changed

However, when using `native_types`:

```php
<?php
declare(strict_types=1);

use native_types;

function main()
{
    $x = 42;
    $x = [1, 2];
}
```

the compiler reports:

```bash
Fatal error: Cannot re-assign `$x` from `php::Array` to `php::Int`
# is not it should be int to array?
```

Does `native_types` need to be enabled to strictly lock the inferred type of numeric variables?

Reference:

https://github.com/swoole/typephp/blob/87c7eda96436ba706111643a88f633273b692e53/src/Parser/TypeConversionTrait.php#L220-L232

When `native_types` is not enabled, numeric types are converted to `php::Var`. This means that a variable initially assigned a numeric value can subsequently be reassigned to a different type:

```php
$x = 42;      // php::Var
$x = [1, 2];  // allowed
```

With `native_types`, the numeric type is preserved:

```php
$x = 42;      // php::Int
$x = [1, 2];  // compile-time error
```

This makes me wonder whether this is the intended behavior. Since variables inferred as `string`, `array`, and `object` cannot be reassigned to a different type, shouldn't numeric variable type inference also strictly preserve the inferred type by default?

Requiring `native_types` for numeric types seems inconsistent with the behavior of other inferred types. Without `native_types`, numeric values are converted to `php::Var`, allowing a variable initially assigned a numeric value to be reassigned to an incompatible type.

jefyokta:
## Description

Fixes #85.

Numeric variables were being converted to `php::Var` when `native_types` was not enabled. This allowed variables initially inferred as numeric types to be reassigned to incompatible types.

For example:

```php
$x = 42;
$x = [1, 2];
```

Previously, this was allowed because `$x` was converted to `php::Var`.

This change preserves the inferred numeric type by default, making its behavior consistent with other inferred types such as `string`, `array`, and `object`.

The reassignment above will now be rejected because `$x` is inferred as `php::Int`.

## Expected behavior

```php
$x = 42;
$x = [1, 2]; // Error
```

`native_types` should no longer be required to strictly preserve the inferred numeric type.

## Implementation

The change is made in `getNativeType()`:

```php
protected function getNativeType(string $type): string
{
    if ($type === Type::INT && $this->bigintTypes) {
        return Type::BIGINT;
    }
    if ($type === Type::FLOAT && $this->decimalTypes) {
        return Type::DECIMAL;
    }
    //-  return $this->nativeTypes ? $type : Type::VAR;
     return $type;
}
```

This method is only called once, so changing this behavior does not introduce additional conversion paths or require broader changes to the type inference logic. the only one line that call it:
https://github.com/swoole/typephp/blob/87c7eda96436ba706111643a88f633273b692e53/src/Parser/AssignOpTrait.php#L650-L652
If the intended design is that strict numeric type inference should explicitly require `native_types`, this PR can be rejected accordingly.

### DIFF
--- a/phpunit/code/re-assign-numeric-without-using-native-types.php
+++ b/phpunit/code/re-assign-numeric-without-using-native-types.php
@@ -1,0 +1,7 @@
+<?php
+
+function main()
+{
+    $x = 1;
+    $x = [];
+};
--- a/phpunit/src/AssignTest.php
+++ b/phpunit/src/AssignTest.php
@@ -9,6 +9,11 @@ class AssignTest extends \BaseTest
         $this->exec('Cannot re-assign `$obj` from `php::Str` to `php::Object`', 're-assign.php');
     }
 
+    public function testReAssignNumericValueWithoutUsingNativeTypes()
+    {
+        $this->exec('Cannot re-assign `$x` from `php::Array` to `php::Int', "re-assign-numeric-without-using-native-types.php");
+    }
+
     public function testAssignClass()
     {
         $this->exec('Cannot re-assign typed object `$obj1` from `stdClass` to `ArrayObject`', 're-assign-2.php');
--- a/src/CompilerBase.php
+++ b/src/CompilerBase.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * This file is part of TypePHP.
  *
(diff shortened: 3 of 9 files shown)


jefyokta:
i forgot to test to build tpc with the new tpc, anyway `$target` convertion problem already fixed, but got incompetible with ForeachIterator's phpx.
ForeachIterator::nextKeyValue doesnt exists in phpx v2.7.0? i found it the  in master branch, but not in release
```
error: no member named 'nextKeyValue' in 'php::ForeachIterator'
   28 |         while (tmp_var_0.nextKeyValue(key, value)) {
      |                ~~~~~~~~~ ^
1 error generated.
Fatal error: compile failed:
#0 [internal function]: TypePhp\Diagnostics\CliDiagnostic
```
https://github.com/swoole/phpx/blob/f5663b75c313d817680d26fcae021d67ef32e4f6/include/phpx.h#L2680

PR #101, opened 2026-09-08
perf(compiler): cache calls with literal string method names

yavon007:
Calls such as `$object->{'run'}()` currently bypass method call caches because the AST method name is a string expression rather than an identifier. The name is nevertheless fixed, so repeated calls redo callable resolution unnecessarily.

Allow literal string method names to use the existing request-local method caches, including nullsafe and scoped calls. Runtime receiver resolution, class/name guards, visibility handling and magic-method fallback remain in place. Variable and computed method names retain their existing uncached paths.

Adds codegen coverage and a runtime regression covering repeated calls, inherited overrides, magic methods, private scope and nullsafe receivers. The standalone benchmark includes reproduction instructions and raw timings.

### Performance

10 million calls through an `object` parameter: median 357.32 ms before versus 235.60 ms after (**34.06% less elapsed time**). All nine alternating measured pairs were faster after a discarded warm-up pair; checksums matched native PHP. Linux ARM64 in Docker, PHP 8.5.10 ZTS, PHPX `db0aabd`, GCC `-O2`, no LTO, baseline TypePHP `91f133f1`.

This is an isolated fixed-name call benchmark on a shared development machine, not an application-wide improvement or a comparison against native PHP execution speed.

### Validation

- `CallCacheCodegenTest`: 3 tests / 21 assertions pass. The new assertion fails with the original MethodCallTrait.
- Three targeted PHPTs pass: `literal-method-cache`, `call-cache-dispatch`, and `call-cache-arguments`.
- Native PHP and compiled benchmark checksums match (`505000000`).
- `git diff --check` passes. Full-suite and cross-platform validation were not run.


matyhtf (commented):
实际项目中几乎不可能这样用，既然方法名使用表达式肯定是变化的，若是常量直接写成名称即可。因此类似的优化，是很鸡肋的，只是在测试中有效。

matyhtf:
To clarify the review above (originally in Chinese):

In real projects this pattern is almost never used: if a method name is written as an expression, it is by definition dynamic; if it is a constant literal, you can simply write the method name directly. So this kind of optimization has little practical value and only helps in benchmarks/tests.

PR #102, opened 2026-09-08
Avoid temporary Variants when binding hoisted numeric property slots

yavon007:
When binding an already-hoistable `$this` int or float property to a native
reference, generated code constructs a temporary `Variant` through `attr()`
and immediately calls `unwrap_ptr()`. Bind directly to the checked object slot
and dereference existing property references instead. Existing hoisting safety
checks, property offsets and write conversions are preserved.

In an isolated benchmark repeatedly updating an int and float property,
300 million calls took a median 1.7556 s before and 1.2549 s after across nine
alternating pairs (28.5% less elapsed time; all nine pairs faster), using
PHP 8.5.10 ZTS on Linux ARM64 with GCC -O2.

Validation: 22 native-property code-generation tests and four targeted PHPTs
pass, including inherited properties with existing external references,
object escape, unset and float access. Performance on other platforms and
NTS has not been measured.

<details>
<summary>Standalone benchmark source</summary>

```php
<?php
class Counter {
    public int $value = 1;
    public float $weight = 0.5;
    public function step(int $delta): int {
        $this->value += $delta;
        $this->weight += 0.5;
        return $this->value;
    }
}
function main(int $argc, array $argv): void {
    $n = $argc > 1 ? (int) $argv[1] : 1000000;
    $counter = new Counter();
    $sum = 0;
    for ($i = 0; $i < $n; ++$i) {
        $sum += $counter->step($i % 3);
    }
    echo $sum, ':', $counter->weight, "\n";
}
```

Build as a binary with `optimize: 2`. Run each binary with argument `300000000`, after a warmup, alternating baseline/candidate order for nine rounds. Verify identical output (`45000000250000000:150000000.5`), zero exit status and empty stderr. Timings include process startup.
</details>

### DIFF
--- a/phpunit/src/NativePropertyTest.php
+++ b/phpunit/src/NativePropertyTest.php
@@ -59,7 +59,7 @@ class NativePropertyTest extends \BaseTest
         }
 
         $code = file_get_contents($outputFile);
-        $this->assertStringContainsString('php::Int &_object_prop_this___value = Z_LVAL_P(this_.attr(', $code);
+        $this->assertStringContainsString('php::Int &_object_prop_this___value = Z_LVAL_P(php::unwrap_zval(OBJ_PROP(this_.checkedObject("Attempt to read property"), ', $code);
         $this->assertStringContainsString('php::Int &_object_prop_box__value = Z_LVAL_P(box.attr(', $code);
         $this->assertStringContainsString('_object_prop_this___value += (2L);', $code);
         $this->assertStringContainsString('_object_prop_box__value += (2L);', $code);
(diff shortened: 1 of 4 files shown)


matyhtf:
Closing this PR because the compiler should not call PHPX’s internal checkedObject() API or emit the raw Zend OBJ_PROP macro directly. The regression test covering inherited properties and existing references has been adopted. The optimization is instead implemented through new public PHPX Object::attrInt() and Object::attrFloat() APIs that return typed C++ references. bool remains on the existing value path and is intentionally not optimized as a reference.

matyhtf:
https://github.com/swoole/typephp/commit/38d828cbbe8f485e0d0d8a241526e110e5962396

PR #103, opened 2026-09-09
feat(closure): closure parameter type narrowing from call-site literals

yuan-dian:
### Summary

Infer closure parameter types from call-site literal arguments to generate native C++ types instead of `php::Var`. When all call sites pass the same literal type, the lambda signature uses the native type directly, eliminating runtime type checks.

### Performance

| Scenario | Before | After | Speedup |
|---|---|---|---|
| `fn($x)(3.14)` float | 28ms | 4ms | **7x** |
| `fn($x)("hello")` string | 140ms | 22ms | **6x** |
| `fn($x)(true)` bool | 10ms | 3ms | **3x** |
| `fn($x)(42)` int | 6ms | 3ms | **2x** |
| `fn($x)($i)` variable | 35ms | 35ms | no change |
| `fn($x)([1,2,3])` array | 360ms | 360ms | no change |

### Changes

**Core Logic:**
- `LocalClosureAnalyzer.php`: Track call sites per candidate, `inferParamTypes()` detects int/float/string/bool/array literals, unary ops, boolean expressions, string concatenation, cast expressions, and ConstFetch
- `ClosureGenerator.php`: Use inferred types in lambda signatures, skip type checks when effectiveType is native, add `newClosureWithParameters` for closures with type checks
- `Translator.php`: Wire `inferParamTypes()` into candidate processing
- `FunctionContext.php`: Document `callSites` and `inferredParamTypes` keys

**Tests:**
- `ClosureParamTypeTest.php`: 22 unit tests covering type declarations, call-site literals, multi-call fallback, unary ops, boolean expressions, string concatenation, cast expressions, goto, nested functions, and class method guard
- `closure-param-type.php`: Test fixture with 20+ scenarios
- `closure-param-type-class.php`: Class method guard test
- `closure-param-type-inference.phpt`: End-to-end integration test

### How It Works

1. **Analysis phase**: `LocalClosureAnalyzer` tracks all call sites for each closure candidate and infers parameter types from literal arguments
2. **Code generation**: If all call sites pass the same type (or type declaration exists), the lambda uses native C++ type (e.g., `php::Int`, `php::Float`) instead of `php::Var`
3. **Fallback**: Multiple call sites with different types → `php::Var` + runtime type check (unchanged behavior)

### Scope

- ✅ Scalar type declarations (int, float, string, bool)
- ✅ Array type declarations
- ✅ Call-site literal inference (all scalar types + array)
- ✅ Unary expressions (-42, +42)
- ✅ Boolean expressions (===, ||, instanceof)
- ✅ String concatenation ("hello" . "world")
- ✅ Cast expressions ((int)"42")
- ✅ ConstFetch (true, false)
- ✅ FuncCall (count, strlen, sizeof)
- ❌ Class method closures (guard: stays as Zend closure)
- ❌ Closures with goto (invalidated)
- ❌ Closures with nested functions (invalidated)

### Safety

- No behavior change for existing code
- Multi-call closures with different types fall back to `php::Var`
- Class method closures remain as Zend closures
- Closures with goto/labels are invalidated
- 60 closure tests pass, full test suite 1987 tests (8 pre-existing failures unchanged)


matyhtf (changes requested):
Thank you for working on closure parameter type inference. The current implementation is not ready to merge yet because it introduces several correctness regressions in TypePHP's strong type semantics.

Please address the following blockers:

1. Explicit parameter declarations must remain authoritative. `resolveEffectiveClosureParamType()` currently always returns the inferred call-site type and ignores `Node\Param::$type`. As a result, `fn(int $value) => $value` called with `"12"` returns a string instead of raising `TypeError`. Call-site inference must never bypass or replace an explicit declared type.

2. Please reuse the compiler's canonical expression-type detection instead of maintaining a second AST-only type system in `LocalClosureAnalyzer::detectArgType()`. The duplicate implementation classifies every float literal as `FLOAT`, while TypePHP promotes high-precision literals to `Decimal`. For example, passing `3.14159265358979323846` currently generates a `php::Float` lambda parameter but a `php::Variant`/Decimal argument, and the generated C++ does not compile. BigInt and other existing promotions need the same canonical handling.

3. Several operator result types are incorrect. The spaceship operator (`<=>`) returns `int`, not `bool`, and unary `+`/`-` cannot simply inherit a boolean operand type. At present both `1 <=> 2` and `-true` can be narrowed to bool and change `-1` into `true`.

4. The PR description says narrowing occurs when all call sites have the same type, but `inferParamTypes()` falls back to `VAR` whenever there is more than one call site. Please either implement the documented behavior safely or adjust the scope and description.

5. Please strengthen the tests with isolated fixtures and observable generated signatures/runtime behavior. The current large fixture and generic substring assertions can pass by matching unrelated closures. In particular, the explicit typed-parameter assertion checks parameter name `x`, while the typed closures use `a`; the goto test also succeeds because another closure already contains `newClosureWithParameters`. Please add negative/regression cases for explicit parameter types, high-precision Decimal literals, `<=>`, unary numeric operators, and multiple same-type call sites.

Suggested direction: keep `LocalClosureAnalyzer` responsible for locality/escape analysis and call-site collection, obtain argument types through the existing compiler type-analysis path, narrow only when every call site has the same canonical compatible native type, and otherwise retain `php::Var` plus the normal runtime type check.

matyhtf (changes requested):
感谢继续完善这个优化。最新版本的基础方向是合理的：为不逃逸的局部 Closure 收集直接调用点，并在调用类型一致时生成强类型 C++ lambda。

不过我基于最新提交 `5ecb725d4` 做了实际 C++ 编译验证，目前仍有两个阻塞问题，需要修改后再合并。

### 1. `detectTypeOfExpr()` 的 PHP 语义类型不等于生成表达式的 C++ ABI 类型

当前 `inferCallSiteArgType()` 直接把 `detectTypeOfExpr()` 的结果用于 lambda 参数类型，但实际生成的 C++ 表达式可能仍然是 `php::Var`。

直接编译本 PR 新增的 `phpunit/code/closure-param-type.php` 即可复现：

```bash
php bin/tpc.php phpunit/code/closure-param-type.php -m lib -o closure_param_type
```

主要错误包括：

```text
could not convert ‘dl1’ from ‘php::Decimal’ to ‘php::Var’
no known conversion from ‘php::Variant’ to ‘php::Decimal’
no known conversion from ‘php::Variant’ to ‘php::Int’
```

其中：

- Decimal 用例将参数收窄为 `php::Decimal`，但调用表达式 `php::toDecimal(...)` 实际返回 `php::Var`；lambda 返回 `dl1` 时也不能自动转换为 `php::Var`。
- `2 ** 3` 的 PHP 结果类型是 `int`，但实际生成的 `php::fn::pow(...)` 返回 `php::Var`，无法直接传给 `php::Int` 参数。

因此不能只依据 PHP 语义类型收窄。建议增加“生成表达式是否具有相同 native C++ 表示”的判断/白名单，或者在调用边界统一执行可靠的拆箱转换。Decimal、BigInt、BigFloat 等类型在转换完整之前应回退到 `php::Var`。

### 2. 显式 `array`、`object` 和具体 class 参数遇到错误实参时会生成无效 C++

例如：

```php
$fn = fn(array $value): int => count($value);
return $fn(42);
```

当前生成：

```cpp
auto fn = [](php::Array value) mutable -> php::Var { ... };
fn(42L);
```

最终由 C++ 编译器报错，而不是保留原来的 PHP 参数检查并在运行时抛出 `TypeError`。`object` 和具体 class 参数也有相同问题。

原因是调用点当前仅实现 `int/float/bool/string` 四种转换；`array/object/class` 没有转换。`object/class` 的类型检查位于 lambda 内部，但错误实参无法进入 lambda；`array` 的检查则因为参数已被标记为 native type 而被跳过。

建议在完整实现非标量调用边界转换前，让这些显式类型继续使用 `php::Var` 参数和现有运行时检查；或者补齐 Array/Object/class 的参数校验与转换。

### 测试需要补充实际 C++ 编译

当前新增 PHPUnit 中的 `compileFixture()` 只执行 PHP → C++ 转换并检查生成字符串，没有编译生成的 C++。因此相关 PHPUnit 和 CI 都通过了，但 PR 自己新增的完整 fixture 实际无法编译。

请至少增加：

1. 完整 fixture 的 C++ 编译/运行测试，特别是 Decimal 和 `pow()`；
2. 显式 `array/object/class` 参数接收错误实参的测试，确认程序能够编译，并产生预期的参数类型错误；
3. 对所有允许收窄的类型验证生成表达式的真实 C++ 表示，而不只是断言 lambda 签名字符串。

完成以上两项语义修复和端到端测试后，再进行下一轮 review。

matyhtf:
PR #104 暂时不可合并。主测试路径正确，但存在一个新增回归和两个未覆盖的同类问题。

 ## 1. [P1] 非精确属性类型被错误用于可见性判定

  addObject($object, $property->class) 会使后续 findNativeMethod() 按属性声明类型解析方法，即使运行时对象可能是子类。

  复现：
```php
  class Base {
      private function value(): string {
          return 'base';
      }
  }

  final class Child extends Base {
      public function value(): string {
          return 'child';
      }
  }

  final class Holder {
      public Base $target;

      public function __construct() {
          $this->target = new Child();
      }

      public function run(): string {
          return $this->target->value();
      }
  }
```
  PHP 以及 PR 基线 a1782233 都输出：

  child

  PR 当前提交却在编译期报错：

  `Method `Base::value()` is not accessible`

  这里属性声明类型只能用于查找参数签名，不能自动视为方法分派的精确类型。建议将“签名解析类型”和“可直接调用的精确类型”分开保存；无法证明具体运行时类时仍走动态分派。

##  2. [P1] interface 类型属性仍然丢失引用参数

  条件只接受：

 ```
$this->hasClass($property->class)
```

  因此 interface 属性被排除：
```php
  interface Recorder {
      public function record(array &$events): void;
  }

  final class Holder {
      public Recorder $target;

      public function run(array &$events): void {
          $this->target->record($events);
      }
  }
```
  实测仍然产生：
```php
  Argument #1 ($events) must be passed by reference, value given
  array(0) {}
```
  interface 同样提供完整的方法签名。它不应该直接反虚化到某个实现，但应使用 interface 签名生成引用参数。

##  3. [P2] static typed property 未覆盖

  当前分支仅处理 Expr\PropertyFetch，没有处理 Expr\StaticPropertyFetch：
```php
  public static Target $target;

  self::$target->record($events);
```
  实测同样出现引用警告，且 $events 写入丢失。

matyhtf:
结论：PR #103 目前仍不可合并。最新提交 8e94301d 存在两个阻塞问题。

##  1. [P1] 类型推断结果不等于生成表达式的 C++ ABI 类型

  inferCallSiteArgType() 根据 detectTypeOfExpr() 将类型属性读取识别为 php::Int，但调用生成阶段可能通过 php::deindirect() 得到 php::Var。

  复现：
```php
  class Box {
      public int $value = 42;
  }

  function main(): void {
      $box = new Box();
      $fn = fn($x) => $x + 1;
      var_dump($fn($box->value));
  }
```
  生成结果：
```cpp
  auto fn = [](php::Int x) mutable -> php::Var { ... };
  tmp_var_2 = fn(php::deindirect(tmp_var_1));
```

  真实 C++ 编译失败：
```
  no known conversion for argument 1 from ‘php::Var’ to ‘php::Int’
```
  这里不能继续增加表达式特例。需要区分：

  - PHP 语义类型；
  - 表达式实际生成的 C++ ABI 类型。

  只有后者确定为原生类型时才能收窄 lambda 参数。

## 2. [P1] 多参数类型检查顺序错误

  parseNativeLocalClosureCall() 把可能抛出异常的转换直接放进 C++ 调用参数：
```
  fn(
      php::toIntArgExact(first, ...),
      php::toIntArgExact(second, ...)
  );
```
  C++17 不保证函数参数从左到右求值。GCC 实测以下代码先报告参数 #2：
```
  $fn = fn(int $first, int $second) => 0;
  $fn("bad-first", "bad-second");
```
  TypePHP 输出：
```
  Argument #2 ($second) must be of type int
```
  PHP 正确行为是首先报告参数 #1。所有可能抛异常的参数转换都应按 PHP 顺序提前写入临时变量，再调用 lambda。

##  3. 测试仍不足以保护这项优化

  PR 的 40 个 PHPUnit 测试、123 个断言全部通过，CI 也全部通过，但测试只检查生成的 C++ 字符串，没有真正编译和执行，因此没有发现上述编译错误和行为差异。

  建议至少增加：

  - typed property 参数的真实 C++ 编译测试；
  - 多个无效参数的运行测试；
  - 对比 PHP 和 TypePHP 的首个 TypeError 参数编号。

matyhtf:
感谢你为 Closure 参数类型推断和性能优化投入的工作，也感谢针对前几轮 review 补充修复和端到端测试。

综合评估后，我们决定不继续推进这项优化，并关闭此 PR。这是实现方向和维护成本上的取舍，而不只是等待当前几个问题修复后再合并。

主要考虑如下：

- 箭头函数和 Closure 在实际项目中常作为回调参数传给其他函数，例如 array_walk、array_map 等。这类场景不属于本 PR 所针对的、可证明不逃逸且直接调用的局部闭包，因此不能直接获得这里展示的参数收窄收益。
- 当前性能示例主要是理想化的局部直接调用用例，尚不足以说明这项优化对实际项目有足够收益。
- 为保持现有行为，需要同时处理 PHP 语义类型与 C++ ABI、实参求值顺序、参数写入、引用传递及动态类型检查。几轮 review 已表明，这部分复杂度和维护成本远超最初预期，而且分析本身还可能增加编译耗时。

因此，我们倾向于让这些动态回调继续通过 PHP Zend Bridge 运行，不引入本 PR 的静态参数类型收窄逻辑。后续若有代表性的真实项目性能瓶颈和数据，再单独评估更小、更明确的优化方向。

再次感谢你的贡献。

matyhtf:
English version of the closing decision:

Thank you for your work on Closure parameter type inference and performance optimization, and for addressing the earlier review feedback with fixes and end-to-end tests.

After considering the trade-offs, we have decided not to pursue this optimization and to close this PR. This is a decision about the implementation direction and maintenance cost, rather than a request to keep fixing the remaining issues before merging.

Our main considerations are:

- In real-world projects, arrow functions and Closures are commonly passed as callbacks to functions such as array_walk and array_map. These uses fall outside the non-escaping, directly invoked local closures targeted by this PR, so they do not directly benefit from the parameter narrowing demonstrated here.
- The current performance examples primarily cover idealized local direct-call scenarios. They do not yet establish sufficient benefits for representative real-world workloads.
- Preserving existing behavior requires handling PHP semantic types versus C++ ABI representations, argument evaluation order, parameter mutation, reference passing, and dynamic type checks. The review iterations have shown that the complexity and maintenance cost are substantially higher than initially expected. The analysis itself can also increase compilation time.

For these dynamic callback scenarios, we prefer to retain execution through the PHP Zend Bridge rather than introduce the static parameter type narrowing proposed in this PR. If representative real-world profiling later identifies a concrete bottleneck, we can evaluate a smaller, more narrowly scoped optimization separately.

Thank you again for your contribution.

PR #104, opened 2026-09-12
fix(compiler): preserve signatures for typed property method calls

yavon007:
# fix(compiler): preserve signatures for typed property method calls

Calls such as `$this->target->record($events)` were lowered dynamically even when
`target` had a non-nullable declared project class. This bypassed the known
parameter signature: an `array &$events` argument emitted a by-reference warning
and lost its writes. It also routed eligible calls through Zend wrappers.

Materialize the property receiver once before argument evaluation, record its
declared object type, and reuse existing direct-call eligibility and argument
lowering. Do not mark the receiver as an exact runtime class; existing override,
abstract-class, visibility and magic-call handling remain in effect. Nullable and
unknown property types retain their current paths.

Validation:

- Baseline runtime fixture produced two reference warnings and an empty events
  array; the candidate matches PHP, including named arguments, receiver replacement
  during argument evaluation, subclass dispatch, nullsafe access, getter invoked
  once, and uninitialized-property errors.
- Targeted PHPUnit: 25 tests, 83 assertions passed.
- New compiled PHPT passed.
- Existing dungeon: 51 behavioral assertions passed with identical output.
- Alternating five-run O2 benchmarks with identical outputs: one million property
  calls, 29.97 → 22.63 ms median (24.5% lower). Normal dungeon, 485.27 → 473.70 ms
  (2.4%); exception-heavy dungeon, 1216.81 → 1226.59 ms (no improvement).

The application-scale performance result is modest; this change primarily restores
known call signatures and removes an unnecessary runtime dispatch where the
existing devirtualizer permits it. Benchmark environment: Linux ARM64 container,
PHP 8.5.10 ZTS, PHPX `4b3a472`, baseline TypePHP `a1782233`.

### DIFF
--- a/benchmark/property-method-call/README.md
+++ b/benchmark/property-method-call/README.md
@@ -1,0 +1,24 @@
+# Typed property method calls
+
+This benchmark calls a final class method through a declared object property
+1,000,000 times, returning checksum `7000000`. It isolates receiver/property and
+method-call overhead; it is not a whole-application performance estimate.
+
+Build from the repository root with a matching PHPX runtime:
+
+```sh
+php bin/tpc.php benchmark/property-method-call/project.yml --no-progress \
+  --build-dir /tmp/typephp-property-method-build -o /tmp/property-method-call
+/tmp/property-method-call
+```
+
+For a before/after comparison, use the same PHPX, compiler flags and workload in
+both checkouts, separate build directories, and alternate the two binaries after
+a warmup. Verify equal output before comparing elapsed time.
+
+An initial Linux ARM64/PHP 8.5.10 ZTS/PHPX `4b3a472` O2 comparison against TypePHP
+`a1782233` measured five-run median process times of 29.97 ms before and 22.63 ms
+after (24.5% lower elapsed time). Timings include process startup. The normal
+22-command dungeon scenario improved only about 2.4%, and its exception-heavy
+28-command scenario showed no reliable improvement. These results demonstrate a
+local call-path improvement, not a claim that all applications become 24.5% faster.
--- a/benchmark/property-method-call/main.php
+++ b/benchmark/property-method-call/main.php
@@ -1,0 +1,22 @@
+<?php
+final class PropertyCounter
+{
+    public function value(): int { return 7; }
+}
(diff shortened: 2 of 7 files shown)


matyhtf (changes requested):
Thanks for working on this. The main fixture compiles and runs correctly, but I found three correctness issues that need to be addressed before merging.

1. A declared property class is not necessarily the receiver's exact runtime class.

`addObject($object, $property->class)` makes `findNativeMethod()` resolve visibility against the declared class before runtime dispatch. This breaks a valid private-method shadowing case:

```php
class Base {
    private function value(): string { return 'base'; }
}

final class Child extends Base {
    public function value(): string { return 'child'; }
}

final class Holder {
    public Base $target;

    public function __construct() {
        $this->target = new Child();
    }

    public function run(): string {
        return $this->target->value();
    }
}
```

PHP and the base commit both print `child`; this PR rejects the program at compile time with:

```text
Method `Base::value()` is not accessible
```

The declared property type should be usable for signature lookup, but it must not become an exact dispatch type. Please separate signature resolution from devirtualization/direct-call eligibility. If the runtime class is not proven exact, this case must retain dynamic dispatch.

2. Interface-typed properties still lose known by-reference signatures.

The new condition accepts `hasClass()` only. For example:

```php
interface Recorder {
    public function record(array &$events): void;
}

final class Holder {
    public Recorder $target;

    public function run(array &$events): void {
        $this->target->record($events);
    }
}
```

The compiled program still reports that `$events` must be passed by reference and the write is lost. An interface cannot provide a direct implementation target, but it does provide the parameter signature and should be used for argument lowering while keeping runtime dispatch.

3. Static typed-property receivers have the same reference-loss bug.

The new path handles only `Expr\PropertyFetch`; a call such as `self::$target->record($events)` through a non-nullable typed static property still emits the by-reference warning and loses the write. Please cover `Expr\StaticPropertyFetch` as well, or explicitly narrow the PR scope and address this before claiming general typed-property method-call support.

Please add compiled runtime regressions for all three cases. The current CI and added tests pass, but they do not exercise these paths.

PR #108, opened 2026-09-12
Resolve function imports before dispatch and policy checks

yavon007:
Calls through `use function` aliases can resolve to the wrong target when the
call's casing differs from the import, or when a same-named compiled global
function exists. For example, a user function imported as GET_CALLED_CLASS
is bypassed by `get_called_class()`, causing the built-in to run and throw.

Normalize function-alias keys during registration and lookup. When an
unqualified call names an imported function, use only the imported target for
native-function resolution rather than checking global functions first.
Keep absolute/qualified calls and class/constant imports unchanged.
Resolve the imported target before special-function and capability-policy
checks as well: safe functions must not be rejected because their alias names
a restricted built-in, and restricted functions must not evade policy through
an innocuous alias.

Add codegen and PHP/AOT behavioral coverage for mixed-case aliases, built-in
aliases, grouped imports, global name collisions, and explicit global calls.
Codegen assertions verify the selected imported and global targets; the
compiled regression matches PHP output exactly with empty stderr.
Additional compile-only tests ensure an imported exec alias is rejected under
both Nano and WASI, while a user function aliased as extract remains callable.

Focused PHPUnit: 115 tests / 632 assertions pass (11 existing deprecations).
The complete 135-file compiler self-build and --version check pass. The
resulting compiler also builds and runs the regression with PHP-identical output.
Tested on Linux ARM64, PHP 8.5.10 ZTS, GCC and PHPX 4b3a472. No benchmark
speedup is claimed.


matyhtf (changes requested):
Thanks for the fix. The regular call paths now resolve `use function` aliases correctly, but the first-class callable path still emits the original alias spelling instead of the resolved target name.

For example:

```php
use function strlen as SIZE;

function main(): void {
    $callable = sIzE(...);
    echo $callable('abc'), "\n";
}
```

PHP prints `3`, but the compiled TypePHP program fails at runtime with:

```text
Failed to create closure from callable: function "sIzE" not found or invalid function name
```

The same failure occurs with an imported compiled user function. In `FunctionCallTrait`, the `genPlaceHolder()` paths still use `identifierToStr($expr->name)` / the original `$placeHolder`, so the resolved name produced by `getNamespacedFuncName()` is lost. Please make statically named placeholder/first-class-callable paths use the fully resolved import target, and add a PHPT regression covering a mixed-case imported first-class callable (preferably both an internal function and a compiled user function).

The existing function-import resolution, collision, and Nano/WASI policy tests pass.

PR #110, opened 2026-09-12
fix(compiler): preserve runtime class dispatch and introspection for polymorphic objects

prateekbhujel:
When an object comes from a function or method return (like `$a = getAnimal()`), SSA marks it in `stableObjects`. But that only tracks the upper-bound declared type, not the actual runtime subclass:

- `get_class($a)` was getting folded at compile time to the base class instead of returning the runtime subclass (and returned a string on `null` instead of throwing `TypeError`)
- `$a::who()` statically resolved to the base class, bypassing subclass overrides

Swapped `isStableObject` with `isset($this->context->exactObjects[...])` in `genGetClassOptimized` and `parseStaticCall`. If the class isn't proven exact via `new ConcreteClass()`, it safely falls back to runtime dispatch (`php::fn::get_class` and `php::callStaticMethod`).

Added unit tests for both exact and polymorphic cases.

### DIFF
--- a/phpunit/code/polymorphic-dispatch.php
+++ b/phpunit/code/polymorphic-dispatch.php
@@ -1,0 +1,46 @@
+<?php
+
+class BaseAnimal
+{
+    public static function identify(): string
+    {
+        return 'base';
+    }
+}
+
+class DogAnimal extends BaseAnimal
+{
+    public static function identify(): string
+    {
+        return 'dog';
+    }
+}
+
+function makeAnimal(): BaseAnimal
+{
+    return new DogAnimal();
+}
+
+function getExactClass(): string
+{
+    $dog = new DogAnimal();
+    return get_class($dog);
+}
+
+function getPolymorphicClass(): string
+{
+    $animal = makeAnimal();
+    return get_class($animal);
+}
+
+function callExactStatic(): string
+{
+    $dog = new DogAnimal();
+    return $dog::identify();
+}
+
+function callPolymorphicStatic(): string
+{
+    $animal = makeAnimal();
+    return $animal::identify();
+}
--- a/phpunit/src/PolymorphicClassDispatchTest.php
+++ b/phpunit/src/PolymorphicClassDispatchTest.php
@@ -1,0 +1,44 @@
+<?php
+
+use TypePhp\CompilerTest;
+
+final class PolymorphicClassDispatchTest extends BaseTest
+{
+    public function testPolymorphicClassIntrospectionAndStaticDispatch(): void
+    {
+        global $translator;
+
+        $compiler = CompilerTest::create(TYPEPHP_ROOT_PATH);
+        $translator = $compiler;
+        $source = TYPEPHP_ROOT_PATH . '/phpunit/code/polymorphic-dispatch.php';
+        $compiler->addFiles([$source]);
+        $compiler->prepareFile($source);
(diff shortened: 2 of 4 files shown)


matyhtf (changes requested):
Thanks for addressing the incorrect devirtualization. Public polymorphic `get_class()` and `$object::method()` dispatch work in the tested cases, but two runtime issues still need to be fixed.

1. The new dynamic `$object::method()` path loses the lexical class scope. A protected static call that is valid in PHP is treated as coming from global scope:

```php
class Base {
    protected static function identify(): string {
        return static::class;
    }

    public static function exercise(): string {
        $object = makeChild(); // declared Base, returns Child
        return $object::identify();
    }
}
```

PHP returns the child class name, while the compiled program throws:

```text
Call to protected method Base::identify() from global scope
```

The runtime dispatch must preserve both the runtime target class and the lexical callable scope. Emitting an unscoped `php::callStaticMethod()` is insufficient here.

2. The PR description says nullable values now produce PHP's `TypeError`, but `get_class()` on a nullable object still segfaults. The non-exact path emits `php::fn::get_class(value)`, whose current fast path dereferences `obj.ce()` without validating null. For example, a `?Base` return containing null causes SIGSEGV instead of a catchable `TypeError`. Please use a checked/runtime path for values that may be null, or make the helper enforce PHP's argument semantics.

The added PHPUnit test only checks generated code text, so neither runtime failure is covered. Please add behavioral PHPT coverage for a protected polymorphic static call and nullable `get_class()`, in addition to the public dispatch cases.

prateekbhujel:
Updated. Dynamic static calls inside methods now use `php::callScoped` to keep caller scope for protected methods, and non-exact `get_class` falls back to runtime `php::call` so null throws TypeError. Added the requested phpt tests for both as well.

Details on the changes:

1. **Static call scope**:
   In `MethodCallTrait::parseStaticCall()`, when dynamic static dispatch occurs inside a method (`$this->methodDef !== null`), we emit `php::callScoped($fn, $this->getCallableScopeExpr(), ...)`. This initializes `CallableScope` with the executing class frame so Zend passes the access check for protected methods, while keeping `$fn` dynamically target-bound (`Child::identify`) so late static binding (`static::class`) resolves the runtime subclass. Top-level/global callers continue using `php::callStaticMethod()`.

2. **Nullable `get_class()` handling**:
   In `FuncCallOptimizer::genGetClassOptimized()`, non-exact objects now return `false` instead of emitting the unchecked C++ helper `php::fn::get_class()`. This routes execution through `genRuntimeFunctionCall()` (`php::call` targeting Zend's `zif_get_class`), which correctly returns the runtime subclass for objects and triggers standard PHP `TypeError` on `null` without segfaulting. Proven exact instances (`new Foo()`) continue to fold at compile time.

3. **Behavioral PHPT coverage**:
   - `tests/compiler/static/polymorphic-static-call-scope.phpt`: verifies protected static method dispatch on polymorphic instances through static and instance caller methods, subclass overrides, and global public calls.
   - `tests/compiler/stdlib/get_class_nullable.phpt`: verifies runtime class introspection on nullable objects returning `null` (catching `TypeError`), literal `get_class(null)`, exact objects, and polymorphic objects.
   - Updated unit assertions in `phpunit/src/PolymorphicClassDispatchTest.php`.

PR #120, opened 2026-09-19
Linux builds gets an rpath

alganet:
Depends on nothing, but found while testing #119 (keeping the
embed build inside its own prefix). That PR makes the private `libphp.so` build
succeed for a non-root user; this is the next thing you hit.

A program built on Linux does not start:

```
$ ./main
./main: error while loading shared libraries: libphpx.so: cannot open shared object file: No such file or directory
```

`libphpx.so` sits beside the phpx checkout, `libphp.so` under the prefix the
libphp installer chose. The link line names both with `-L`, which tells the
linker where to look and the loader nothing:

```
g++ ... -L'.../vendor/swoole/phpx/lib' -L'/home/user/.typephp/lib' -lphpx -lphp ...
```

```
$ readelf -d main | grep -E 'RPATH|RUNPATH'
(nothing)
```

### Cause

The mechanism is already present and already reached on Linux.
`NativeCommandOptionsTrait` asks the platform for `getDefaultRpaths()` on every
non-nano build, and `GccLikeBackend` emits `-Wl,-rpath` for each entry. Only the
list is empty:

| | `getDefaultRpaths()` |
| --- | --- |
| `PlatformBase` | `[]` |
| `Macos` | phpx lib dir + `resolvePhpLibDir($phpDir)` |
| `Android`, `Ios` | `[]`, explicitly |
| `Linux`, `UnixPlatform`, `Wasi` | no override |

`Linux` is the only native platform inheriting the empty default.

### Change

`Macos`'s override moves up to `UnixPlatform`. The directories are the ones the
`-L` flags already name, each still guarded by `is_dir()`, so macOS behaves as
before. `Wasi` gains the explicit `[]` that `Android` and `Ios` already have,
so the move does not hand a cross target host paths.

### Result

```
$ readelf -d main | grep RUNPATH
Library runpath: [/home/user/typephp/vendor/swoole/phpx/lib:/home/user/.typephp/lib]

$ env -i ./main
$ echo $?
0
```

Before, the same program under `env -i` exits 127 with the loader error above.

### Scope

The recorded paths are absolute, matching what macOS has always emitted, so the
program runs where it was built. Relocatable output wants `$ORIGIN` with the
libraries shipped alongside, or `--full-static`; neither is decided here.


matyhtf (approved):
The implementation looks correct and fixes the Linux runtime loader failure.

I verified it locally by building a Linux executable. The resulting ELF contains both the PHPX and PHP library directories in `RUNPATH`, and it starts successfully under `env -i` without relying on `LD_LIBRARY_PATH`. The platform/backend test suite also passes (65 tests, 205 assertions), as do the PHP syntax checks and `git diff --check`.

Non-blocking suggestion: please consider adding a regression test that creates temporary PHPX/PHP library directories and verifies that Linux returns both as default rpaths, while WASI still returns an empty list. The current platform tests do not exercise the new Linux behavior.

